PR #5, opened 2016-10-31
`fush` -> `flush` 💪

twokul:


rwjblue:
But what about fushing?

rwjblue:
Thank you!

Issue #1, opened 2016-10-31
Automatically reopen primary classes to add mixin.


rwjblue:
Specifically, mix `ember-lifeline/mixins/run` into the following classes:

* `Ember.Component`
* `Ember.Route`
* `Ember.Controller`
* `Ember.Helper`
* `Ember.Service`

And mix `ember-lifeline/mixins/dom` into:

* `Ember.Component`

scalvert:
I don't believe we have plans to do this. Please feel free to reopen if I'm mistaken.

PR #11, opened 2016-12-08
Mention use of Mixins in README

simonihmig:
I just wondered why all the methods were `undefined`. Thought that the addon would reopen `Ember.Component` and so on to add them automatically, but apparently that is not what is happening (until #1 lands). But having to apply the mixins by yourself was nowhere mentioned, also the examples were misleading in that regard. 

Added the missing pieces!

Super useful addon btw! :)


rwjblue:
Awesome, thank you!

simonihmig:
Wow, that was quick! :) You're welcome, keep up the great work for the ember community!  👏   😃

PR #13, opened 2017-01-17
Add removeEventListener method to DomMixin

simonihmig:
I had the need to remove an event listener during the active lifecycle of a component. This can happen especially for high volume events like `scroll` or `mousemove`, where you maybe want to listen to them conditionally, not all the time.

This adds the matching `removeEventListener` method to the `DomMixin`.

Added a test just for one simple case (with and without the `passive` flag), not all the different cases you have provided for `addEventListener`. Not sure if that makes sense to duplicate all of them also for `removeEventListener`? If you want me to do that, just tell me!

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,17 @@ Passivity refers to the ability for a handler to either `preventDefault` or,
 in our implementation, `stopPropagation`. In dev mode and testing, use of
 these APIs on the event object will raise an exception.
 
+### `removeEventListener`
+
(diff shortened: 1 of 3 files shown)


simonihmig:
@rwjblue had a chance to look at this?

rwjblue:
Awesome! I'll try to release in the AM (I'm in computer free time for the night 😎).

offirgolan:
ping @rwjblue. Can we get a new release with this?

rwjblue:
Released in 1.0.4

PR #14, opened 2017-02-14
[FEATURE] Schedule Task

### LINKED ISSUE #2: Enable usage of `Ember.run.schedule`
This could either be via options passed to `runTask`, or another top level method specifically for this (i.e. `scheduleTask`).

As of https://github.com/emberjs/ember.js/pull/14550 (will be included in 2.8.3, 2.9.1, 2.10.0-beta.3+), it is possible to cancel things scheduled into a specific queue via `Ember.run.schedule`. In a sample application `run.schedule` is one of the primary remaining use cases of `Ember.run` that still requires manual `this.isDestroying` checks.

offirgolan:
Resolves #2.

@rwjblue this is just a quick implementation with some test cases. I want to get your feedback first before I continue. 

- [x] `scheduleTask` implementation
- [x] Test cases
- [ ] How to handle Ember < 2.8
- [ ] Documentation

### DIFF
--- a/addon/mixins/run.js
+++ b/addon/mixins/run.js
@@ -92,6 +92,54 @@ export default Mixin.create({
   },
 
   /**
+   Adds the provided function to the named queue to be executed at the end of the RunLoop.
+   The timer is properly canceled if the object is destroyed before it is invoked.
+
+   Example:
+
+   ```js
+   import Component from 'ember-component';
+   import ContextBoundTasksMixin from 'web-client/mixins/context-bound-tasks';
+
+   export default Component.extend(ContextBoundTasksMixin, {
+     init() {
+       this._super(...arguments);
+       this.scheduleTask('afterRender', () => {
+         console.log('This runs at the end of the run loop (via the afterRender queue) if this component is still displayed');
+       })
+     }
+   });
+   ```
+
+   @method scheduleTask
+   @param { Function } callback the callback to run at the provided time
+   @param { ...* } args arguments to pass to the callback
+   @param { Number } [timeout=0] the time in the future to run the callback
+   @public
+   */
+  scheduleTask(queue, callbackOrName, ...args) {
+    assert(`Called \`scheduleTask\` without a string as the first argument on ${this}.`, typeof queue === 'string');
+    assert(`Called \`scheduleTask\` on destroyed object: ${this}.`, !this.isDestroyed);
+    let type = typeof callbackOrName;
+
+    let cancelId = run.schedule(queue, this, () => {
+      let cancelIndex = this._pendingTimers.indexOf(cancelId);
+      this._pendingTimers.splice(cancelIndex, 1);
+
+      if (type === 'function') {
+        callbackOrName.call(this, ...args);
+      } else if (type === 'string' && this[callbackOrName]) {
+        this[callbackOrName](...args);
+      } else {
+        throw new Error('You must pass a callback function or method name to `scheduleTask`.');
+      }
+    });
+
+    this._pendingTimers.push(cancelId);
+    return cancelId;
+  },
+
+  /**
    Runs the function with the provided name after the timeout has expired on the last
    invocation. The timer is properly canceled if the object is destroyed before it is
    invoked.
--- a/tests/unit/mixins/run-test.js
+++ b/tests/unit/mixins/run-test.js
@@ -124,6 +124,86 @@ test('runTask tasks can be canceled', function(assert) {
   }, 10);
 });
 
+test('scheduleTask invokes async tasks', function(assert) {
+  assert.expect(2);
+
+  let subject = this.subject();
+  let done = assert.async();
+  let hasRun = false;
+
+  run(() => {
+    subject.scheduleTask('actions', () => {
+      hasRun = true;
+      assert.ok(true, 'callback was called');
+      done();
+    });
+
+    assert.notOk(hasRun, 'callback should not have run yet');
+  });
+});
+
+test('scheduleTask invokes named functions as async tasks', function(assert) {
+  assert.expect(4);
+  let done = assert.async();
+  let subject = this.subject({
+    run(name) {
+      hasRun = true;
+      assert.equal(this, subject, 'context is correct');
+      assert.equal(name, 'foo', 'passed arguments are correct');
+      assert.ok(true, 'callback was called');
+      done();
+    }
+  });
+  let hasRun = false;
+
+  run(() => {
+    subject.scheduleTask('sync', 'run', 'foo');
+    assert.notOk(hasRun, 'callback should not have run yet');
+  });
+});
+
+test('cancels tasks added with `scheduleTask`', function(assert) {
+  assert.expect(2);
+  let subject = this.subject();
+  let done = assert.async();
+  let hasRun = false;
+
+  run(() => {
+    subject.scheduleTask('actions', () => {
+      hasRun = true;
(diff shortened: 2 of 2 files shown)


rwjblue:
@mixonic - Mind reviewing for API thoughts here?

@offirgolan:

> How to handle Ember < 2.8

We could feature detect and monkey patch 😝.

```js
(function() {
  let cancelationToken = Ember.run.schedule('afterRender', () => {});
  if (cancelationToken === undefined) {
    Ember.run.schedule = function(/* queue, target, method */) {
      Ember.assert(
       `You have turned on testing mode, which disabled the run-loop's autorun. ` +
        `You will need to wrap any code with asynchronous side-effects in a run`,
        Ember.run.currentRunLoop || !Ember.testing
      );

      return Ember.run.backburner.schedule(...arguments);
    };
  }
})();
```

offirgolan:
@rwjblue that works for me. Should we make this into a separate polyfill addon since this is something that will be need by machty/ember-concurrency#121?

addon/mixins/run.js
```diff
@@ -92,6 +92,54 @@ export default Mixin.create({
...
+   @public
+   */
+  scheduleTask(queue, callbackOrName, ...args) {
+    assert(`Called \`scheduleTask\` without a string as the first argument on ${this}.`, typeof queue === 'string');
+    assert(`Called \`scheduleTask\` on destroyed object: ${this}.`, !this.isDestroyed);
+    let type = typeof callbackOrName;
+
+    let cancelId = run.schedule(queue, this, () => {
+      let cancelIndex = this._pendingTimers.indexOf(cancelId);
+      this._pendingTimers.splice(cancelIndex, 1);
```
mixonic: In this case I wonder if a lazy no-op of the scheduled function might be a better performance option than explicit cancellation. Explicit cancellation means walking the list of scheduled tasks to find the cancelled one, then modifying the queue. A lazy no-op would simply mean the scheduled task sits on the queue until it run, then it determines it doesn't need to and truncates.

tests/unit/mixins/run-test.js
```diff
@@ -124,6 +124,86 @@ test('runTask tasks can be canceled', function(assert) {
...
+  let hasRun = false;
+
+  run(() => {
+    subject.scheduleTask('actions', () => {
+      hasRun = true;
+      assert.ok(true, 'callback was called');
+      done();
+    });
+
+    assert.notOk(hasRun, 'callback should not have run yet');
```
mixonic: I think you mean for this assertion to be outside the `run`? You want to assert that it hasn't run at the end of the runloop as well.

tests/unit/mixins/run-test.js
```diff
@@ -124,6 +124,86 @@ test('runTask tasks can be canceled', function(assert) {
...
+      assert.ok(true, 'callback was called');
+      done();
+    }
+  });
+  let hasRun = false;
+
+  run(() => {
+    subject.scheduleTask('sync', 'run', 'foo');
+    assert.notOk(hasRun, 'callback should not have run yet');
+  });
```
mixonic: Similarly I think you want to assert `hasRun` is true after the runloop here.

tests/unit/mixins/run-test.js
```diff
@@ -124,6 +124,86 @@ test('runTask tasks can be canceled', function(assert) {
...
+    subject.scheduleTask('actions', () => {
+      hasRun = true;
+      assert.ok(false, 'callback was called');
+    });
+
+    assert.notOk(hasRun, 'callback should not have run yet');
+    run(subject, 'destroy');
+  });
+
+  window.setTimeout(() => {
```
mixonic: the `setTimeout` calls here and in the next test seem a bit paranoid. Confirming the code hasn't run after the `run` call would be sufficient.

addon/mixins/run.js
offirgolan: Not sure I follow you here. Would you care to give an example?

tests/unit/mixins/run-test.js
offirgolan: This assert makes sure that the scheduled task did not occur instantly after executing `subject.scheduleTask`.

tests/unit/mixins/run-test.js
offirgolan: Same as my previous comment.

tests/unit/mixins/run-test.js
offirgolan: This line is taken from the `runTask` tests. Just following what was already there.

offirgolan: I see your point though. Since the run loop should complete by this line then there is no need for it. I'll make the change 😄

tests/unit/mixins/run-test.js
mixonic: I'm suggesting an assertion for `assert.ok(hasRun` after the runloop, as that would show when the task was supposed to run (during the runloop queue flush). As written, this test doesn't assert timing beyond a) not immediately and b) eventually. I think you want to also assert c) after the runloop `scheduleTask` is called in. At the least. You could go further and assert what queue it runs in, but that seems ok to pass on to me.

addon/mixins/run.js
mixonic: `run.cancel` needs to do *something*, right? So it walks the list of all scheduled tasks (which may be long) and then must allocate a new array of valid tasks or mutate the existing array to remove the cancelled item.

Instead, it can be cheaper to allow a function to execute as scheduled, and simply make it a no-op at execution time. I looked at this approach for the `runTask` code, but I think in those cases avoiding the macro task queue item of a timeout makes cancellation worth the browser's cancellation expense. Here, I think the tradeoff is a bit different. Firing the scheduled task as a no-op is very cheap, cheaper than an extra iteration and allocation (or mutation) for cancellation.

The difference is between explicit cancellation:

```javascript
let tasks = [];
let task = { id: '123', fn: () => { /* do something */ } };
let cancelledTaskId = task.id;

tasks.push(task);

/* Now to cancel we must iterate and allocate a new `tasks` */
tasks = tasks.reject(t => t.id === cancelledTaskId);

/* Flush */
tasks.forEach(t => t.fn());
```

And lazy no-op:

```javascript
let tasks = [];
let taskCancelled = false;
let task = {
  id: '123',
  fn: () => {
    if (!taskCancelled) {
      /* do something */
    }
  }
};

tasks.push(task);

/* Now to cancel we change the value of the cancellation flag */
taskCancelled = true;

/* Flush */
tasks.forEach(t => t.fn());
```

Different implementations, the tests should be the same though.

offirgolan: @mixonic thank you for taking the time to answer my question. I understand the differences between the implementations but what I dont understand is how to actually implement a no-op scheduled task when we are solely dependent on Ember's backburner to handle all the scheduling / cancellations. Are you imposing we create a `cancelTask` api to do something like this? There has to be something that toggles the `taskCancelled` property.... 

There are also times that you want to cancel a scheduled task prematurely. How would you do this in a no-op way?

tests/unit/mixins/run-test.js
offirgolan: Oh I see what you mean here. I will go ahead and add that to the test cases 😄

nickiaconis:
I have use for this, and it seems like something that'd be useful to others as well. What's the current status? Can I help in any way?

rwjblue:
Likely needs a rebase at least...

offirgolan:
Woot woot!

Issue #17, opened 2017-02-23
destruct Mixin, assert, run from Ember


foxnewsnetwork:
Suggestion
===
In the preamble to https://github.com/rwjblue/ember-lifeline/blob/master/addon/mixins/run.js, we see:

```javascript
import Ember from 'ember';
import Mixin from 'ember-metal/mixin';
import run from 'ember-runloop';
import { assert } from 'ember-metal/utils';
```
Please consider changing to:

```javascript
import Ember from 'ember';
const { Mixin, run, assert } = Ember;
```

Why
===
While this isn't a big deal to most ember users in the world of 2.x.x world, a lot of enterprise applications of Ember are still stuck in the 1.x.x world where `ember-metal` and `ember-runloop` aren't a thing yet. I know it's our own fault for having forefathers whom have sinned and written cancer code that now keeps us stuck in EoL version of Ember... 

But haven't we suffered enough? So our forefathers sinned; so we ignored the best practices of DDAU, spammed `controller` everywhere, and have `window.setInterval` peppered all over our app. But in order for us to move forward, we need addons like this to fix our problems. And the fact that we who need it the most can't access it is quite a painful irony.

Please, make the change.

Also, I can make PR and write up a test if you guys are too busy

Workaround
===
But what if @rwjblue has no mercy for us sinners? Here's what we must do to work-around (aka atone):

https://gist.github.com/foxnewsnetwork/71ed4b5c973e0c2249b80dd421e99acd

rwjblue:
Happy to review a PR

foxnewsnetwork:
Thank you, here is pr https://github.com/rwjblue/ember-lifeline/pull/18

rwjblue:
Closed by #18

PR #18, opened 2017-02-23
destruct from ember instead of depending on ember-cli-shims

foxnewsnetwork:
Background
===
Right now, it's impossible for ember apps stuck in the dead world of 1.x.x to use this addon. This is because https://github.com/rwjblue/ember-lifeline/blob/master/addon/mixins/run.js#L2 imports from `ember-metal` and `ember-runloop` instead of destructuring from ember

Changes
===
Change the import statements to:
```javascript
import Ember from 'ember';
const { Mixin, run, assert } = Ember;
```

Alternatives
===
As @rwjblue points out in slack

> Those imports come from ember-cli-shims

which means any addon that also installs https://github.com/ember-cli/ember-cli-shims will be able to use this addon out-of-thebox.

Therefore, we can have the following two alternatives:

1. Installation blueprint on this addon `blueprints/ember-lifeline/index.js` that installs `ember-cli-shims`
2. Write in the README.markdown somewhere that the user will need to auto-manually install `ember-cli-shims` or write their own shims for `ember-metal` and `ember-runloop` 

Reference
===
https://github.com/rwjblue/ember-lifeline/issues/17

### DIFF
--- a/addon/mixins/run.js
+++ b/addon/mixins/run.js
@@ -1,7 +1,10 @@
 import Ember from 'ember';
-import Mixin from 'ember-metal/mixin';
-import run from 'ember-runloop';
-import { assert } from 'ember-metal/utils';
+
+const {
+  Mixin,
+  run,
+  assert
+} = Ember;
 
 let _shouldPollOverride;
 function shouldPoll() {


rwjblue:
FWIW, ember-cli-shims is included out of the box by ember-cli (and has been since very close to ember-cli's inception).

This change seems totally reasonable to me though. Thanks for the PR!

foxnewsnetwork:
Hooray, thank you! BTW, I tried using `ember-cli-shims` just now and the following was what I got:

screen shot 2017-02-23 at 4 14 34 pm

Alas, the depths of the sins committed by my code forefathers at my company are more grave than I imagined; not only are we stuck on an older version of `ember`, we're also stuck on an older version of `ember-cli`

PR #19, opened 2017-02-27
Allowing for wrapping of callbacks

alexander-alvarez:
As per discussion over yonder https://github.com/alphasights/ember-scrollable/pull/57

It would be nice if there was a way to block until the handlers of async events (scroll, wheel) have fired. There's no native way to get access to the listeners on an object, but using `ember-lifeline` as an abstraction we can have access to them, wrap them, and insert them back during testing.

My way of solving this initially (just to spur conversation) was to move most of the body of `addCoalescedEventListener` (since the core functionality was static) out of the class, export it, and make the coalesced function return a resolved promise after all handlers have been called.

Open to anything that lets me axe https://github.com/alphasights/ember-scrollable/blob/master/tests/integration/components/scroll-content-element-test.js#L133..L149

### DIFF
--- a/addon/mixins/dom.js
+++ b/addon/mixins/dom.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 
 const {
   Mixin,
+  RSVP,
   $,
   run,
   merge,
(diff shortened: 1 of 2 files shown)


alexander-alvarez:
@rwjblue  thoughts on this?

I don't need all of the removal / readding approach, but it would be nice to be able to wrap around the execution of these event listeners if not at least for testing purposes.

PR #20, opened 2017-03-17
Remove references to web-client in docs

code0100fun:
Looks like some restructuring happened and the docs were outdated.


rwjblue:
Thank you!

PR #21, opened 2017-03-20
[Bugfix] - Adding assertions to ensure _super has been called

scalvert:
If `_super` has not been called within the init chain for classes/mixins utilizing the `ContextBoundTasksMixin` and the `ContextBoundEventListenersMixin` the calls to `willDestroy` will error out due to the members within those mixins not being initialized. This PR adds assertions to identify those classes.

**Reviewers**:
@rwjblue 

**Changes**:
- Added assertions in both run.js and dom.js to identify classes that haven't called `_super` in `init`.


scalvert:
@rwjblue ping?

stefanpenner:
even better :)

PR #25, opened 2017-06-07
Fix scope of coalesced event listeners

simonihmig:
While the scope of event handlers was correctly bound (`run.bind`) to the component instance adding the listener for non-passive listeners, for coalesced/passive listeners (on the same DOM element, e.g. `body`) this was not the case. So when using `this` in a listener, this could be a completely different component instance! (fwiw probably the first component calling `_addCoalescedEventListener` and thus attaching the `coalescedCallback` for that given element).

This fixes it and adds a corresponding test!

### DIFF
--- a/addon/mixins/dom.js
+++ b/addon/mixins/dom.js
@@ -151,7 +151,8 @@ export default Mixin.create({
     }
   },
 
-  _addCoalescedEventListener(element, eventName, callback) {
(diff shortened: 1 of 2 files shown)


rwjblue:
@simonihmig - I just updated to resolve some conflicts, can you double check?

I think this is good to go...

simonihmig:
@rwjblue LGTM!

PR #27, opened 2017-06-15
Add assertion if `runTask` is invoked but `init` was not correctly run

stefanpenner:


rwjblue:
Thank you for working on this! I'm going to close in favor of #21 (has tests, and covers more usages).

PR #28, opened 2017-06-15
Moving arrays to be lazy-allocated

scalvert:
As the description suggestions, if this mixin is applied its arrays are currently being eagerly allocated. This PR makes them lazy.

@rwjblue @stefanpenner

### DIFF
--- a/addon/mixins/dom.js
+++ b/addon/mixins/dom.js
@@ -84,10 +84,6 @@ function removeHandlerFromListenerData(handler) {
  @public
  */
 export default Mixin.create({
-  init() {
-    this._super(...arguments);
-  },
-
   /**
    Attaches an event listener that will automatically be removed when the host
    object is dropped from DOM.
@@ -116,7 +112,6 @@ export default Mixin.create({
   addEventListener(selector, eventName, callback, _options) {
     assert('Must provide an element (not a jQuery selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
     assert('Called addEventListener before the component was rendered', this._currentState === this._states.inDOM);
-    assert(`Called \`addEventListener\` without \`init\` calling \`_super\` in ${this}. Please ensure \`init\` is calling \`_super\`.`, typeof this._listeners !== 'undefined');
 
     // Ember.assign would be better here, but Ember < 2.5 doesn't have that :(
     let options = merge(merge({}, DEFAULT_LISTENER_OPTIONS), _options);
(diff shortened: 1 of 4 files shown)


addon/mixins/dom.js
```diff
@@ -86,9 +86,16 @@ function removeHandlerFromListenerData(handler) {
 export default Mixin.create({
   init() {
     this._super(...arguments);
+  },
+
+  get _listeners() {
```
rwjblue: discussed briefly with @scalvert about this, we can't use real JS getters here for two reasons:

* it creates the getter on the POJO passed in to `Mixin.create()` which isn't what we want
* it makes it impossible to do the correct cleanup in `willDestroy` without creating the array during teardown

I proposed an alternative approach which would use a specific API for pushing things into the array:

```js
_trackListener(item) { 
  let listeners = this._listeners;
  if (!listeners) { 
    listeners = this._listeners = []; 
  }

  listeners.push(item);
}
```

scalvert: Realized this after Rob mentioned it. Foolish mistake. Fixed.

addon/mixins/dom.js
```diff
@@ -84,13 +84,6 @@ function removeHandlerFromListenerData(handler) {
  @public
  */
 export default Mixin.create({
-  init() {
-    this._super(...arguments);
-
-    this._listeners = [];
```
rwjblue: Sorry, I know I mentioned to remove the `init` but I forgot that we need to shape the object properly also. 

Can you bring this section back, but instead of assigning these to a new empty array we need to assign to undefined.

addon/mixins/run.js
```diff
@@ -41,14 +41,6 @@ export function pollTaskFor(label) {
  @public
  */
 export default Mixin.create({
-  init() {
-    this._super(...arguments);
-
-    this._pendingTimers = [];
```
rwjblue: Same as above, we need to shape the object by setting these properties to undefined.

addon/mixins/run.js
```diff
@@ -274,16 +267,22 @@ export default Mixin.create({
...
     this._super(...arguments);
 
-    assert(`Called \`willDestroy\` without \`init\` calling \`_super\` in ${this}. Please ensure \`init\` is calling \`_super\`.`, typeof this._pendingTimers !== 'undefined');
-
     cancelTimers(this._pendingTimers);
     cancelDebounces(this._pendingDebounces);
     clearPollers(this._pollerLabels);
+  },
+
+  getOrAllocateArray(propertyName) {
```
rwjblue: This method should be underscored

addon/mixins/dom.js
scalvert: Ah you're correct. Will do.

addon/mixins/run.js
scalvert: Putting back and slotting props.

addon/mixins/run.js
scalvert: Agree.

PR #29, opened 2017-06-16
Document the full method signature for debounceTask and throttleTask

jsturgis:
This change helps to avoid confusion about what order the method args need to be
in for debounceTask and throttleTask.


rwjblue:
Thank you!

PR #40, opened 2017-07-27
Removing dependency on jQuery for ContextBoundEventListenersMixin

scalvert:
@rwjblue do you feel like this should be tested against a version of Ember w/o jquery?


scalvert:
Will fix it up once I figure out what I broke when pulling the thread to upgrade the addon to the latest blueprint.

addon/mixins/dom.js
```diff
@@ -67,12 +62,27 @@ function removeHandlerFromListenerData(handler) {
...
   }
 }
 
+function getOrCreateProperty(element, name) {
+  let value = element[name];
+
+  if (!value) {
+    value = {};
+
+    Object.defineProperty(element, name, {
```
stefanpenner: defineProperty here is likely going to be a performance hit

scalvert: Fixed!

scalvert:
@rwjblue thoughts?

addon/mixins/dom.js
```diff
@@ -110,43 +110,53 @@ export default Mixin.create({
...
    */
   addEventListener(selector, eventName, callback, _options) {
-    assert('Must provide an element (not a jQuery selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
+    assert('Must provide an element (not a DOM selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
     assert('Called addEventListener before the component was rendered', this._currentState === this._states.inDOM);
 
     // Ember.assign would be better here, but Ember < 2.5 doesn't have that :(
     let options = merge(merge({}, DEFAULT_LISTENER_OPTIONS), _options);
     let element = findElement(this.element, selector);
+    let eventNames = eventName.split(' ');
```
rwjblue: I think we should drop this with the jquery removal. Specifying the event like `oninput change` is a very "jquery" way of doing it. If we intend to be a thin shim over native API's we should remove this.

addon/mixins/dom.js
```diff
@@ -110,43 +110,53 @@ export default Mixin.create({
...
    @param { Function } _callback the callback to run for that event
    @public
    */
   removeEventListener(selector, eventName, callback, _options) {
-    assert('Must provide an element (not a jQuery selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
+    assert('Must provide an element (not a DOM selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
 
     let options = merge(merge({}, DEFAULT_LISTENER_OPTIONS), _options);
     let element = findElement(this.element, selector);
+    let eventNames = eventName.split(' ');
```
rwjblue: Remove (same as above)

addon/mixins/dom.js
scalvert: I'd tend to agree. I'll also drop the corresponding test.

addon/mixins/dom.js
scalvert: 👍

addon/mixins/dom.js
```diff
@@ -110,13 +110,13 @@ export default Mixin.create({
...
    @param { String } _eventName the event name to listen for
    @param { Function } _callback the callback to run for that event
    @public
    */
   addEventListener(selector, eventName, callback, _options) {
-    assert('Must provide an element (not a jQuery selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
+    assert('Must provide an element (not a DOM selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
     assert('Called addEventListener before the component was rendered', this._currentState === this._states.inDOM);
 
     // Ember.assign would be better here, but Ember < 2.5 doesn't have that :(
```
rwjblue: Since we are bumping major versions, and dropping Ember 2.4 support we should be able to use `Ember.assign` and kill this comment (and the `merge(first, merge(second, third))` BS).

addon/mixins/dom.js
```diff
@@ -184,7 +184,7 @@ export default Mixin.create({
       /*
        * Attach the listener and cache the listener for teardown
        */
-      element.on(eventName, coalescedCallback);
+      element.addEventListener(eventName, coalescedCallback);
```
rwjblue: I believe this should indicate that it is passive (thats the only way we get down this path):

```js
element.addEventListener(eventName, coalescedCallback, { passive: true });
```

And I think the corresponding `element.removeEventListener` needs to pass `{ passive: true }` also (inside `_removeCoalescedEventListener`).

rwjblue: Unfortunately, all browsers don't support an options POJO for the last arg to `addEventListener` so we may have to feature detect (one time at the top of the module).  MDN has an example of how to detect:

```js
const PASSIVE_SUPPORTED = (() => {
  let ret = false;

  try {
    let options = Object.defineProperty({}, "passive", {
      get: function() {
        ret = true;
      }
    });

    window.addEventListener("test", null, options);
  } catch(err) {}
  return ret;
})();
```

PR #46, opened 2017-09-30
Adding ability to register arbitrary disposables

scalvert:
Adds two new public methods:

`registerDisposable` - registers an arbitrary function, which can be run on demand or automatically on `destroy`

`runDisposable` - runs a disposable immediately by providing a label identifier

This allows a consistent way to register work for destruction when using external libraries. This is akin to the [IDisposable](https://msdn.microsoft.com/en-us/library/system.idisposable(v=vs.110).aspx) interface, which when implemented allows for the system to dispose of unmanaged resources.


README.md
```diff
@@ -372,10 +372,116 @@ A couple of helpful assertions are provided with the `pollTask` functionality:
...
+
+  respondToDOMEvent() {
+    // do something
+  }
+});
+```
+
+### `runDisposable`
+
+**tl;dr call `this.runDisposable(label)` when you want to explicity run the disposable function without waiting for the object's destruction.**
```
rwjblue: Let’s call the argument “token”

rwjblue: Or “disposableToken”

README.md
```diff
@@ -372,10 +372,116 @@ A couple of helpful assertions are provided with the `pollTask` functionality:
...
+Use `registerDisposable` as a replacement for explictly unbinding any external bindings. A disposable is a function that disposes of bindings that are outside of Ember's lifecyle. This essentially means you can register a function that you want to run to automatically tear down any bindings when the Ember object is destroyed.
+
+It's common to see code written to explicitly unbind event handlers from external libraries.
+
+```js
+// app/components/foo-bar.js
+import Ember from 'ember';
+import DOMish from 'some-external-library';
+
+const { run } = Ember;
```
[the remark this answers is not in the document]
scalvert: 👀

scalvert:
Fixed up `label` -> `token`, @rwjblue.

README.md
```diff
@@ -372,10 +372,116 @@ A couple of helpful assertions are provided with the `pollTask` functionality:
 * A given `label` can only be used once. If the same `label` is used a second time, an error will be thrown.
 * If nothing has been queued for the given label, calling `pollTaskFor(label)` will trigger an error.
 
+### `registerDisposable`
+
+**tl;dr call `this.registerDisposable(fn)` on any component, route, or service to register a function you want to run when the object is destroyed.**
+
+Use `registerDisposable` as a replacement for explictly unbinding any external bindings. A disposable is a function that disposes of bindings that are outside of Ember's lifecyle. This essentially means you can register a function that you want to run to automatically tear down any bindings when the Ember object is destroyed.
```
rwjblue: Can we keep the description focused on generically running some function upon destroy?  This kind of event binding is only one type of thing where this general functionality is useful (another would be websocket subscriptions, yet another might be interacting with web workers, etc).  It seems great to have these sorts of examples, but the actual purpose of `registerDisposable` is to be super generic...

scalvert: Yep, can do.

addon/mixins/run.js
```diff
@@ -367,12 +368,107 @@ export default Mixin.create({
...
+   @public
+   */
+  registerDisposable(disposable) {
+    if (typeof disposable !== 'function') {
+      throw new Error('You must pass a function as a disposable');
+    }
+
+    let disposables = this._getOrAllocateArray('_registeredDisposables');
+
+    return disposables.push(disposable) - 1;
```
rwjblue: Chatted with @scalvert a bit about this in slack.  I think we need to modify so that we track the specific items in the token:

```js
registerDisposable(callback) {
  let disposable = {
    disposed: false,
    callback
  };
  
  let disposables = this._getOrAllocateArray('_registeredDisposables');
  disposables.push(disposable);
  return disposable;
},

runDisposable(disposable) {
  if (disposable.disposed) { return; }
  disposable.disposed = true;  
  disposable.callback();
},
```

scalvert: 👍  on it

addon/mixins/disposable.js
```diff
@@ -0,0 +1,144 @@
...
+  @public
+  */
+  runDisposable(token) {
+    runDisposable(this._registeredDisposables, token);
+  },
+
+  willDestroy() {
+    this._super(...arguments);
+
+    runDisposables(this._registeredDisposables);
```
rwjblue: Generally speaking, the `willDestroy` should be the _inverse_ of setup work.  This means that the default pattern should be:

```js
willDestroy() {
  // do custom cleanup work

  // then call super afterwards
  this._super(...arguments);
}
```

addon/mixins/run.js
```diff
@@ -77,7 +78,7 @@ export default Mixin.create({
     assert(`Called \`runTask\` on destroyed object: ${this}.`, !this.isDestroyed);
 
     let type = typeof callbackOrName;
-    let pendingTimers = this._getOrAllocateArray('_pendingTimers');
+    let pendingTimers = getOrAllocate(this, '_pendingTimers', []);
```
rwjblue: I totally like this new more functional util style, but we have to be careful here.  This particular implementation creates a new array on _every_ invocation (regardless if it is needed).

We should probably just have two util methods (one is `getOrAllocateArray` and the other is `getOrAllocateObject` if we need it) to avoid the needless extra array creation...

testem.js
```diff
@@ -8,12 +8,12 @@ module.exports = {
...
   ],
-  'browser_args': {
-      'Chrome': [
-        '--disable-gpu',
-        '--headless',
-        '--remote-debugging-port=9222',
-        '--window-size=1440,900'
-      ],
-  }
+  // 'browser_args': {
```
rwjblue: @trentmwillis helping us out yet again: https://twitter.com/trentmwillis/status/878277832278802432

Basically, we can make it headless only on CI...

scalvert:
@rwjblue 

Changed to using `disposable` types returned from the `registerDisposable` method. Also ensured that disposables can only be run once. Fixed the eager allocation on `getOrAllocate`, which essentially had nullified the usefulness of this method as you pointed out.

PR #49, opened 2017-10-26
Cleanup throttled functions in willDestroy

sparrowj:
Throttled functions aren't getting cleaned up in willDestroy, while other timers and debounced methods are being cleaned up.

### DIFF
--- a/addon/mixins/run.js
+++ b/addon/mixins/run.js
@@ -208,6 +208,10 @@ export default Mixin.create({
     cancelDebounce(this._pendingDebounces, name);
   },
 
+  cancelThrottle(cancelId) {
+    cancelThrottle(cancelId);
+  },
+
   /**
    Runs the function with the provided name immediately, and only once in the time window
    specified by the timeout argument.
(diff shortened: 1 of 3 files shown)


addon/mixins/run.js
```diff
@@ -239,7 +243,13 @@ export default Mixin.create({
     assert(`Called \`this.throttleTask('${name}', ${timeout})\` where 'this.${name}' is not a function.`, typeof this[name] === 'function');
     assert(`Called \`throttleTask\` on destroyed object: ${this}.`, !this.isDestroyed);
 
-    run.throttle(this, name, timeout);
+    let pendingThrottles = getOrAllocate(this, '_pendingThrottles', Array);
```
chrisrng: Can we add `_pendingThrottles` to init similar to `this._pendingDebounces = undefined;` on L46

addon/mixins/run.js
```diff
@@ -208,6 +208,10 @@ export default Mixin.create({
     cancelDebounce(this._pendingDebounces, name);
   },
 
+  cancelThrottle(cancelId) {
```
chrisrng: Thoughts on making this `cancelThrottle` the same as `cancelDebounce` where you can do something like `this.cancelDebounce('logMe');` ?

sparrowj: So, cancelTimers uses an array, cancelDebounce uses the named object. I prefer how timers works because a) it's simpler and b) it matches the API of run.later. E.g const cancelId = Ember.run.later(fn, 100); Ember.run.cancel(cancelId)

addon/mixins/run.js
sparrowj: Done.

addon/mixins/run.js
chrisrng: Gotcha thanks for the clear up!

PR #51, opened 2017-11-09
Allow calling addEventListener from non component entities (services,…

jackson-dean:
- add documentation for using instance variables to pass information to the throttle callback
- Update assert logic in addEventListener to support calling from non-component entities
- Fix a subtle bug in the previous assert logic whereby a null tagName property would pass the test for existing tagName when it should not.
- Also discovered that the tagName property was not being set automagically on the test component instances for some reason
- Add a new test to make sure the assert logic now properly accounts for non-component entities using the dom mixin

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,30 @@ In this example, the first click will update `time`, but clicks after that
 for 500ms will be disregarded. Then, the next click will fire and start
 a timeout window of its own.
 
+Often it is desired to pass additional arguments to the throttle callback. We
+also need to reference the same function in order for throttling to work. In
+order to acheive this it is recommended to make use of instance variables. This
+enables the throttle function to use the arguments in the state they are in
+at the time the callback is executed:
+
+```js
+import Ember from 'ember';
+import RunMixin from 'ember-lifeline/mixins/run';
+
+const { Component } = Ember;
+
+export default Component.extend(RunMixin, {
+  click(evt) {
+    this._evt = evt;
+    this.throttleTask('updateClickedEl', 500);
+  },
+  updateClickedEl() {
+    this.set('lastClickedEl', this._evt.target);
+  }
+});
+```
+
+
 ### `pollTask`
 
 **tl;dr call `this.pollTask(fn, label)` on any component, route, or service to setup
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ export default Component.extend(RunMixin, {
   },
   updateClickedEl() {
     this.set('lastClickedEl', this._evt.target);
+    this._evt = null;
   }
 });
 ```
--- a/addon/mixins/dom.js
+++ b/addon/mixins/dom.js
@@ -61,8 +61,14 @@ export default Mixin.create({
    @public
    */
   addEventListener(selector, eventName, _callback, options) {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -218,6 +218,30 @@ In this example, the first click will update `time`, but clicks after that
...
+
+const { Component } = Ember;
+
+export default Component.extend(RunMixin, {
+  click(evt) {
+    this._evt = evt;
+    this.throttleTask('updateClickedEl', 500);
+  },
+  updateClickedEl() {
+    this.set('lastClickedEl', this._evt.target);
```
stefanpenner: this should likely also unset `this._evt` `this._evt = null`

addon/mixins/dom.js
```diff
@@ -61,8 +61,14 @@ export default Mixin.create({
    @public
    */
   addEventListener(selector, eventName, _callback, options) {
-    assert('Must provide an element (not a DOM selector) when using addEventListener in a tagless component.', this.tagName !== '' || typeof selector !== 'string');
-    assert('Called addEventListener before the component was rendered', this._currentState === this._states.inDOM);
+    assert(
```
stefanpenner: @rwjblue the idea here is to loosen the asserts, if the raw element is passed in. This allows other parts of the system (not components) to attach listeners to elements and have them be maintained by lifeline.

rwjblue: Seems fine but we more tests around the assertions now.

tests/unit/mixins/dom-test.js
```diff
@@ -17,6 +18,7 @@ moduleForComponent('ember-lifeline/mixins/dom', {
 
     this.owner = getOwner(this);
     this.owner.register(name, Component.extend(ContextBoundEventListenersMixin, {
+      tagName: 'div',
```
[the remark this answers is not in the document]
jackson-dean: That is a good question. Maybe you can answer it better than I. When i dumped the component instance locally it was "null" and the assert checking this.tagName !== '' was passing in cases where null was passed which is the opposite of what we wanted.

addon/mixins/dom.js
rwjblue: These assertion changes are not quite correct, because they required that you add `tagName: 'div'` to the component above.

This specific assertion is only concerned with the `this.tagName === ''` case when a string was provided. The criteria now requires that `this.tagName` is _always_ present (which is not the case by default).  Since the tests are passing now, this indicates that we are missing coverage because they should be failing. 😸

tests/unit/mixins/dom-test.js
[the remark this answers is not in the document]
rwjblue: The default value for `Ember.Component.prototype.tagName` is `null`. If `component.tagName` is `null` when we are rendering, we default to `div` otherwise we use the user provided string.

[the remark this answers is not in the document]
jackson-dean: yeah that is the part I was missing. I wasn't familiar with the mechanism used to set the tagName to div and i was expecting to see "div" in cases where it was not explicitly set. So my change and the assertions were based on incorrect assumption about what to expect. Thanks, makes it much less confusing now!

addon/mixins/dom.js
jackson-dean: Yep adding that "div" was done because of a misunderstanding on my part about what it means to see `null` for tagName. I need to go back and rework these now that I have a full picture.

PR #52, opened 2017-11-15
Add willDestroy implementation to dom mixin

jackson-dean:
- implement willDestroy
- add test to make sure listener gets removed from service instance

### DIFF
--- a/addon/mixins/dom.js
+++ b/addon/mixins/dom.js
@@ -89,7 +89,7 @@ export default Mixin.create({
     let element = findElement(this.element, selector);
     let callback = run.bind(this, _callback);
 
-    if (!this._listeners) {
+    if (!Array.isArray(this._listeners)) {
       this._listeners = [];
     }
 
@@ -113,7 +113,7 @@ export default Mixin.create({
 
     let element = findElement(this.element, selector);
 
-    if (!this._listeners) {
+    if (!Array.isArray(this._listeners)) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


addon/mixins/dom.js
```diff
@@ -141,7 +141,22 @@ export default Mixin.create({
...
   willDestroyElement() {
     this._super(...arguments);
+    this._removeEventListeners();
+    // Determines whether or not we need to perform cleanup in the willDestroy hook
+    this._hasCalledWillDestroyElement = true;
+  },
 
+  willDestroy() {
+    this._super(...arguments);
+    if (!this._hasCalledWillDestroyElement) {
```
rwjblue: I think we can just call _removeEventListeners again, and it will essentially do nothing...

addon/mixins/dom.js
```diff
@@ -141,14 +141,25 @@ export default Mixin.create({
 
   willDestroyElement() {
```
krisselden: remove willDestroyElement, if the component called didInsertElement, willDestroy will be called after willDestroyElement, and the remove closes over the element.

addon/mixins/dom.js
```diff
@@ -141,14 +141,25 @@ export default Mixin.create({
...
+   */
+  _removeEventListeners() {
     if (this._listeners) {
       /* Drop non-passive event listeners */
       for (let i = 0; i < this._listeners.length; i++) {
         let { element, eventName, callback, options } = this._listeners[i];
         element.removeEventListener(eventName, callback, options);
       }
-      this._listeners.length = 0;
+      delete this._listeners;
```
krisselden: just set back to undefined

addon/mixins/dom.js
```diff
@@ -141,14 +141,25 @@ export default Mixin.create({
...
+  willDestroy() {
+    this._super(...arguments);
+    this._removeEventListeners();
+  },
 
+  /**
+   @private
+   */
+  _removeEventListeners() {
     if (this._listeners) {
```
krisselden: check !== undefined or Array.isArray(), don't use falsy, this is internal we control how it is initialized checked.

PR #59, opened 2018-01-09
Converting pollTask to use tokens instead of labels.

### LINKED ISSUE #36: task labels seem problematic
https://github.com/rwjblue/ember-lifeline/blob/master/addon/mixins/run.js#L22-L24

These labels are global, this means that (for example) two components using the same labels will interfere with each other.

If these labels are used, they should most likely be scoped to the `this` in question, and not made global, or sad interleavings will occure.

I also believe these label methods should return explicit tokens to still explicitly cancel. Which would address the above described issues.

scalvert:
Fixes #36 
  
- [ ] ensure major version number incremented as this represents a breaking API change

### DIFF
--- a/addon/mixins/run.js
+++ b/addon/mixins/run.js
@@ -3,6 +3,7 @@ import { run } from '@ember/runloop';
 import { assert } from '@ember/debug';
 import Ember from 'ember';
 import getOrAllocate from '../utils/get-or-allocate';
+import getNextToken from '../utils/get-next-token';
 
 let _shouldPollOverride;
 function shouldPoll() {
@@ -19,12 +20,12 @@ export function setShouldPoll(callback) {
 }
 
 let queuedPollTasks = Object.create(null);
-let pollTaskLabels = Object.create(null);
-export function pollTaskFor(label) {
-  assert(`A pollTask with a label of '${label}' was not found.`, pollTaskLabels[label]);
-  assert(`You cannot advance a pollTask (\`${label}\`) when \`next\` has not been called.`, !!queuedPollTasks[label]);
+let pollTaskTokens = Object.create(null);
+export function pollTaskFor(token) {
+  assert(`A pollTask with a token of '${token}' was not found.`, pollTaskTokens[token]);
+  assert(`You cannot advance pollTask '${token}' when \`next\` has not been called.`, !!queuedPollTasks[token]);
 
-  return run.join(null, queuedPollTasks[label]);
+  return run.join(null, queuedPollTasks[token]);
 }
 
 /**
@@ -45,7 +46,7 @@ export default Mixin.create({
     this._pendingTimers = undefined;
     this._pendingDebounces = undefined;
     this._pendingThrottles = undefined;
-    this._pollerLabels = undefined;
+    this._pollerTokens = undefined;
   },
 
   /**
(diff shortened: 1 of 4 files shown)


addon/mixins/run.js
```diff
@@ -319,12 +320,14 @@ export default Mixin.create({
...
-       this.pollTask((next) => {
+       let token = this.pollTask((next) => {
          this.get('api').request('get', 'some/path')
            .then(() => {
              this.runTask(next, 1800);
            })
-       }, 'foo-bar#watch-some-path');
+       });
+
+       this.set('pollToken', token);
```
scalvert: This is a problem. Looking at the fix now.

stefanpenner:
```
this._pollToken = token;
```

* should likely be private
* should likely not be change tracking aware

scalvert: Agree. Just adhering to the convention of using `set` always. Can change, as it's likely not destined for a template.

stefanpenner: ya, its likely safe for us to teach "private internal state" shouldn't ever be externally observable so it shouldn't ever use `set`

addon/mixins/run.js
```diff
@@ -449,9 +449,9 @@ function cancelTimer(cancelId) {
   run.cancel(cancelId);
 }
 
-function cancelPoll(label) {
-  pollTaskLabels[label] = undefined;
-  queuedPollTasks[label] = undefined;
+function cancelPoll(token) {
+  pollTaskTokens[token] = undefined;
```
stefanpenner: i think we may want to `delete` the properties, otherwise the `pollTaskToken` and `queuePollTask` property list will still grow unbounded.

scalvert: When you say delete, do you mean splice out the item?

stefanpenner: ah this is an array not an object, strange... But ya, if an array is appropriate then we should consider splicing/replacing.

Is there a reason this is an array and not an Object?

scalvert: Actually you were correct in the first place. It's a dictionary. We should delete the prop. I can do that.

addon/mixins/run.js
```diff
@@ -371,25 +373,23 @@ export default Mixin.create({
...
+    let token = getNextToken();
 
-    if (label) {
-      assert(`The label provided to \`pollTask\` must be unique. \`${label}\` has already been registered.`, !pollTaskLabels[label]);
-      pollTaskLabels[label] = true;
+    pollTaskTokens[token] = true;
 
-      getOrAllocate(this, '_pollerLabels', Array).push(label);
-    }
+    getOrAllocate(this, '_pollerTokens', Array).push(token);
```
stefanpenner: do we ever remove the `token` from this `_pollerTokens` array?

scalvert: We never remove the token from the array, but the tokens are used only right before the object's destruction, in `willDestroy`. I suspect we don't need to explicitly remove the token from the `_pollerTokens` array, since it won't be used post destruction.

scalvert:
@rwjblue and I had an offline conversation about the API change to `pollTask`, and additionally the fact that we've made the tokens internal, meaning they're internally generated and returned in that method. This makes testing a little more awkward, as rather than supplying an explicit label, we're now requiring end-users to stash the returned token on the instance, and retrieve it from there.

As a compromise, and also to avoid the need to do a full version bump, I'm going to continue to support the `label` parameter (renaming to `token`), and default to `getNextToken`. This will allow end-users to provide a specific token if they want, but also leverage the new API that generates the token for you if they choose to.

cc: @stefanpenner

addon/mixins/run.js
scalvert: I'm going to look at removing this in my next refactor, which will include moving us over to using weakmaps for managing the task related metadata.

PR #60, opened 2018-01-15
General cleanup and simplification

scalvert:
Cleaning up method signatures and unifying getting tasks by function or string name

### DIFF
--- a/addon/mixins/run.js
+++ b/addon/mixins/run.js
@@ -69,7 +69,10 @@ export default Mixin.create({
    ```
 
    @method runTask
-   @param { Function } taskOrName the task (or name of task) to run at the provided time
+   @param { Function | String } taskOrName a function representing the task, or string
+                                           specifying a property representing the task,
+                                           which is run at the provided time specified
+                                           by timeout
    @param { Number } [timeout=0] the time in the future to run the task
    @public
    */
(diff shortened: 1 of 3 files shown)


addon/mixins/run.js
```diff
@@ -444,6 +459,21 @@ export function cancelBoundTasks(tasks, cancelFn) {
...
+export function getTask(instance, taskOrName, taskName) {
+  let type = typeof taskOrName;
+  let task;
+
+  if (type === 'function') {
+    task = taskOrName;
+  } else if (type === 'string' && instance[taskOrName]) {
+    task = instance[taskOrName];
+  } else {
+    throw new Error(`You must pass a task function or method name to '${taskName}'.`);
```
stefanpenner: nit: `TypeError`

scalvert:
It was a toss up for me too, but we had a consistency problem with our naming. When I started the cleanup it struck me that it made sense to begin to use the language of `tasks`, as the API is all about executing tasks. I'm happy to have a conversation about it, but I do believe this applies some consistency.

scalvert:
Yes, I agree with both you and Stef. I do think that this library is aimed at providing an API that is for scheduling various types of tasks, and as such using consistent terminology is a benefit. I do, however, feel that `task` implies to me a type, not a function. That said, that's what the docs are for :)

PR #61, opened 2018-01-16
Adds Prettier configuration

scalvert:
Adds prettier, including:

1. All required dependencies
1. Precommit hooks
1. Prettier `--write` of all files in addon, app, config, and tests

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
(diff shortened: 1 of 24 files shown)


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
rwjblue: Can you move this config into `.eslintrc.js`?  (can have it in both places, but having it only here means eslint doens't do quite the right thing)

scalvert: Sure. I had it there before, but opted for an rc. I can swap it back.

stefanpenner:
nice!

PR #63, opened 2018-01-19
Refactor to Weakmaps Part 1 - Runtask to disposable

scalvert:
This PR represents the first part of porting lifeline to leveraging `WeakMap` instead of instance arrays. These changes lay the groundwork for `*Task` functions to use `registerDisposable` to manage their lifecycle state.

TODO:
- [x] Port tests from mixin to util
- [x] Convert `runTask` to use `registerDisposable` instead of instance array
- [x] Make tests pass


addon/mixins/disposable.js
```diff
@@ -57,46 +56,6 @@ export default Mixin.create({
...
-
-    let disposables = getOrAllocate(this, '_registeredDisposables', Array);
-    let disposable = toDisposable(dispose);
-
-    disposables.push(disposable);
-
-    return disposable;
-  },
-
-  willDestroy() {
```
rwjblue: IMHO, we should still implement `willDestroy` in the mixin so we can avoid adding a new (and unique) function to each instance...

addon/utils/disposable.js
```diff
@@ -0,0 +1,60 @@
...
+
+function _setupWillDestroy(obj) {
+  if (!obj.willDestroy) {
+    return;
+  }
+
+  if (!obj.willDestroy.patched && obj._registeredDisposables) {
+    let originalWillDestroy = obj.willDestroy;
+
+    obj.willDestroy = function() {
```
rwjblue: This is fine as a last resort, but we don't want to have to do it for the happy path...

scalvert: If we don't do this, we put the onus on the implementer to ensure they correctly dispose of tasks. I don't think that's what we want. 

I agree that monkey patching `willDestroy` will be sub-optimal, but am not sure what the best way of doing this is.

Thoughts?

addon/mixins/disposable.js
scalvert: So you think each mixin should call `runDisposables` themselves?

addon/mixins/run.js
```diff
@@ -88,18 +89,15 @@ export default Mixin.create({
       !this.isDestroyed
     );
 
-    let pendingTimers = getOrAllocate(this, '_pendingTimers', Array);
-
     let cancelId = run.later(() => {
-      let cancelIndex = pendingTimers.indexOf(cancelId);
-      pendingTimers.splice(cancelIndex, 1);
+      getOrAllocate(this, '_registeredDisposables', Map).delete(cancelId);
```
stefanpenner: we can't delete from something not yet allocated, we also wont try to delete something not created, so we can likely avoid this whole `getOrAllocate` during this code path, and instead interact with the map directly, as it WILL be there

addon/mixins/run.js
```diff
@@ -88,18 +89,15 @@ export default Mixin.create({
...
 
       let task = getTask(this, taskOrName, 'runTask');
 
       task.call(this);
     }, timeout);
 
-    pendingTimers.push(cancelId);
+    registerDisposable(this, cancelId, () => run.cancel(cancelId));
```
stefanpenner: I don't think we want a callback registered per task run, per instance. I would like to avoid the overhead of all those extra functions.

We should:

* continue to build up the array per task run
* register a single "cleanup" per instance, that handles all cleanup.

scalvert: I hear you. There effectively is a single cleanup - the map is iterated and cancels are scheduled - but there definitely are multiple callbacks. Are you proposing that we just capture the `cancelId`s, and merely iterate/run.cancel all at once?

addon/mixins/run.js
scalvert: You're right in that we know the map will be there. I can probably just go ahead and remove that `getOrAllocate` call.

addon/mixins/run.js
stefanpenner: > Are you proposing that we just capture the cancelIds, and merely iterate/run.cancel all at once?

If possible yes, that is unless there is a constraint I am missing.

scalvert: I'm leaning towards agreeing with you. It _feels_ wrong to be creating so many atomic functions that essentially do the same thing. I'm going to explore other approaches to simplify.

addon/mixins/disposable.js
scalvert: Done. I've added the `willDestroy`s back in and marked the method as patched.

addon/mixins/run.js
```diff
@@ -88,18 +98,27 @@ export default Mixin.create({
       !this.isDestroyed
     );
 
-    let pendingTimers = getOrAllocate(this, '_pendingTimers', Array);
+    let registeredTimers = getPendingTimers(this);
+
+    if (!this._isRunTaskDisposableRegistered) {
+      registerDisposable(this, () => {
+        for (let i = 0; i < registeredTimers.length; i++) {
```
scalvert: So I think this will close over `registeredTimers`, which may cause a leak. Thoughts @stefanpenner @rwjblue?

scalvert:
@stefanpenner to address your (correct) comments, I've reimplemented this to be a single `disposable` that runs a set of timers to `dispose`. Additionally, I've moved further to our goal by converting our instance arrays that were lazy allocated to `WeakMap`s. This moves us closer towards a functional API and to leveraging storing state in a safe, disconnected fashion.

addon/mixins/disposable.js
```diff
@@ -57,46 +62,12 @@ export default Mixin.create({
...
-
-    disposables.push(disposable);
-
-    return disposable;
+    return registerDisposable(this, dispose);
   },
 
   willDestroy() {
-    runDisposables(this._registeredDisposables);
+    runDisposables(registeredDisposables.get(this));
```
rwjblue: Can we modify this to only need `runDisposables(this)`?  Basically, I think its a _super minor_ leak of the abstraction that `runDisposables` needs direct access to the weakmap here...

addon/mixins/disposable.js
```diff
@@ -14,6 +17,8 @@ export default Mixin.create({
     this._super(...arguments);
 
     this._registeredDisposables = undefined;
+
+    this.willDestroy.patched = true;
```
rwjblue: Due to how Ember does super wrapping, if this mixin is mixed into anything that _also_ implements a `willDestroy` this flag will be lost.  Consider this rough analogue (which is basically what [Ember does here](https://github.com/emberjs/ember.js/blob/7c93445601cb9c917fbb64e80fd85e2f6ead0dbd/packages/ember-utils/lib/super.js#L53-L68)):

```js
function _wrap(func, superFunc) {
  function superWrapper() {
    let orig = this._super;
    this._super = superFunc;
    let ret = func.apply(this, arguments);
    this._super = orig;
    return ret;
  }

  return superWrapper;
}
```

So as you can see, when Ember wraps any other `willDestroy`'s this flag on the func will be lost.

I'd suggest that instead we move to using a private branding on the mixin, that we can check for.

Perhaps, something like:

```js
const DO_CLEANUP_YO = typeof Symbol === 'function' ? Symbol('ember-lifeline cleanup') : Math.floor(Math.random() * new Date());

export default Mixin.create({
  [DO_CLEANUP_YO]: true,

  /* other stuff here */
})
```

addon/utils/disposable.js
```diff
@@ -0,0 +1,86 @@
+import { assert } from '@ember/debug';
+
+export let registeredDisposables = new WeakMap();
+
+let registeredTimers = new WeakMap();
```
rwjblue: I think this should ultimately move to where we define the functional `runTask` method, in the interim lets move it into `addon/mixins/run-task.js` as it will simplify logic there quite a bit...

addon/mixins/run.js
```diff
@@ -53,6 +59,10 @@ export default Mixin.create({
     this._pendingDebounces = undefined;
     this._pendingThrottles = undefined;
     this._pollerTokens = undefined;
+
+    this._isRunTaskDisposableRegistered = false;
```
rwjblue: Hmm, I don't think we need this specific flag on the instance. If we move the weakmap from `getPendingTimers` into this file, we can just check it below with a `.has` check.

addon/mixins/disposable.js
scalvert: This did occur to me, and I wasn't too keep on keeping flags on the object as properties. I like the symbol approach.

addon/mixins/disposable.js
scalvert: Yes. We can totally do this.

addon/mixins/run.js
scalvert: 👍

addon/utils/disposable.js
scalvert: 👍

addon/mixins/disposable.js
scalvert: Wouldn't we do:

```js
export default Mixin.create({
  init() {
    this.[DO_CLEANUP_YO] = true,
  }
  /* other stuff here */
})
```
To set on the instance, not the prototype?

scalvert: NVM, worked this out offline.

addon/mixins/disposable.js
```diff
@@ -10,11 +10,7 @@ import getOrAllocate from '../utils/get-or-allocate';
  @public
  */
 export default Mixin.create({
-  init() {
-    this._super(...arguments);
-
-    this._registeredDisposables = undefined;
-  },
+  [WILL_DESTROY_PATCHED]: true,
```
stefanpenner: this doesn't seem used, lets remove.

addon/mixins/run.js
```diff
@@ -4,6 +4,19 @@ import { assert } from '@ember/debug';
...
+const { WeakMap } = Ember;
+
+/**
+ * A map of instances/timers that allows us to
+ * store cancelIds for scheduled timers per instance.
+ *
+ * @private
+ *
+ */
+let registeredTimers = new WeakMap();
```
stefanpenner: `const`

addon/utils/disposable.js
```diff
@@ -0,0 +1,89 @@
...
+
+const { WeakMap } = Ember;
+
+/**
+ * A map of instances/array of disposables. Only exported for
+ * testing purposes.
+ *
+ * @public
+ */
+export let registeredDisposables = new WeakMap();
```
stefanpenner: `const`

addon/utils/disposable.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+export function runDisposables(obj) {
+  let disposables = registeredDisposables.get(obj);
+
+  if (!disposables) {
+    return;
+  }
+
+  for (let i = 0, l = disposables.length; i < l; i++) {
+    let disposable = disposables.pop();
```
stefanpenner: we should likely not mutate the array while we are iterating it. That will force extra bounds checks for no reason.
We should not also not both caching length.

The following  would be both performant and safe.

```js
let disposables  = registeredDisposables.get(obj);
registeredDisposables.delete(obj); // or set to `null`

for (let i =0; i<disposables.length; i++) {
  disposables[i].dispose(); // or simplified, without a disposable wrapper object `disposables[i]();`
}
```

addon/utils/disposable.js
```diff
@@ -0,0 +1,89 @@
...
+  if (!arr) {
+    registeredDisposables.set(obj, (arr = []));
+  }
+
+  return arr;
+}
+
+function _toDisposable(doDispose) {
+  return {
+    dispose() {
```
stefanpenner: this wrapper object doesn't seem to provide much value, combined with my above suggestions for iteration change. We could simply store `doDispose` in the array directly, no need to allocate more stuff.

addon/utils/flags.js
```diff
@@ -0,0 +1,5 @@
+function getFlag(str) {
```
stefanpenner: i think this file can be dropped

addon/utils/disposable.js
```diff
@@ -0,0 +1,89 @@
...
+export function registerDisposable(obj, dispose) {
+  assert(
+    'Called `registerDisposable` where `obj` is not an object',
+    typeof obj === 'object'
+  );
+  assert(
+    'Called `registerDisposable` where `dispose` is not a function',
+    typeof dispose === 'function'
+  );
+  assert(
```
stefanpenner: this seems to just check that we are running `registerDisposable` on the mixin, if that is the intent we should test that directly. If not, then this seems somewhat confusing unless the functional style has a "setup disposable" in the constructor

addon/mixins/disposable.js
scalvert: It's used in the assertion. https://github.com/rwjblue/ember-lifeline/pull/63/files#diff-69582adbea9112eef4525140eef26a3bR38

addon/mixins/run.js
scalvert: 👍

addon/utils/disposable.js
scalvert: 👍

addon/utils/disposable.js
scalvert: Length was cached due to its mutation during the loop. I could also have used a decremented while loop, which would have iterated from end to start (and I believe would be more performant).

> That will force extra bounds checks for no reason.

Not sure what this statement means. 

I can definitely simplify this by not popping.

addon/utils/disposable.js
scalvert: I agree, and had been thinking about discussing with @rwjblue as to it's refactoring/simplifying.

addon/utils/flags.js
scalvert: Agree. We were initially looking to use this in a few places, but it turns out we didn't need to reuse the functionality.

scalvert: So to be clear, I need the constant to be shared, but the function can go. I use the constant in a few places, and need the symbol to be consistent.

addon/mixins/disposable.js
scalvert: OK, so to be specific here, I believe we still need and want this assertion. The intent of this library is to ensure that we're registering work, and correctly disposing of said work when an object is 'destroyed'. The property as symbol approach is aimed at ensuring that we have correctly implemented the call to `runDisposables(this`, which currently is the way we'll be disposing of work.

This property isn't so much intended to be used to ensure the use of the mixin, but instead used as a way to signal to the library that we've correctly implemented `runDisposables` in the destruction of the object.

I hope this makes more sense.

addon/mixins/run.js
scalvert: Done

addon/utils/disposable.js
scalvert: Done

addon/utils/disposable.js
scalvert: See my [comment](https://github.com/rwjblue/ember-lifeline/pull/63#discussion_r166052030) above. This isn't checking we're running `registerDisposable`, but checking that _any_ class or component is calling `runDisposables`, which ensures that tasks are correctly canceled. 

I believe we still need and want this.

addon/utils/disposable.js
scalvert: Agree. I've simplified, with an intent to make this API private in a backwards incompatible breaking change. Ultimately, this will help us converge towards whatever 'cleanup' mechanism is implemented in Ember itself, which will allow us to change the underlying implementation of `registerDisposable` to use that mechanism instead.

addon/utils/flags.js
scalvert: Removed the `getFlag` function and just inlined the exported const.

scalvert:
@stefanpenner any chance you can take another look at this?

PR #64, opened 2018-01-23
use flat array for `_listeners`

bekzod:


addon/mixins/dom.js
```diff
@@ -133,29 +133,33 @@ export default Mixin.create({
...
 
   willDestroy() {
     this._super(...arguments);
-    if (Array.isArray(this._listeners)) {
+    if (this._listeners !== undefined) {
       /* Drop non-passive event listeners */
-      for (let i = 0; i < this._listeners.length; i++) {
-        let { element, eventName, callback, options } = this._listeners[i];
+      for (let i = 0; i < this._listeners.length; i += 5) {
+        let element = this._listeners[i];
```
rwjblue: Lets make these use a constant to make it super clear what we are doing, I'm thinking something like:

```js
let element = this._listeners[i + ELEMENT_INDEX];
let eventName = this._listeners[i + EVENT_NAME_INDEX];
let callback = this._listeners[i + CALLBACK_INDEX];
let options = this._listeners[i + OPTIONS_INDEX];
```

What do you think?

scalvert:
Thanks for this PR, @bekzod! I like the optimization here. As @rwjblue mentioned, let's try to see if we can make this super readable and clear for the next reader. I know I'll likely forget what is going on here in a month if I don't keep looking at it.

addon/mixins/dom.js
```diff
@@ -133,29 +133,33 @@ export default Mixin.create({
     }
 
     // We cannot use Array.findIndex as we cannot rely on babel/polyfill being present
-    for (let i = 0; i < this._listeners.length; i++) {
-      let listener = this._listeners[i];
+    for (let i = 0; i < this._listeners.length; i += 5) {
```
scalvert: I'd even add a constant to represent the magic 5, something like `LISTENER_INFO_SIZE`. This will help hint to people that the indexes represent a grouping of information, and the array is not merely a contiguous set of disjointed items.

addon/mixins/dom.js
bekzod: yep, nice idea will update PR and will make same in bb

scalvert:
We're close here. There's just a prettier error for dangling comma that needs fixing and we're good!

PR #65, opened 2018-02-05
document using this.element with addEventListener

steveszc:
This would have saved me a few minutes of head scratching when upgrading from the old jQuery-compatible API (`this.addEventListener(this.$(), 'scroll', fn);`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,6 +659,9 @@ this.addEventListener('.someClass', 'scroll', fn);
 // Attach to a DOM node
 this.addEventListener(document.body, 'click', fn);
 
+// Attach to a component's root element
+this.addEventListener(this.element, 'scroll', fn);
+
 // Attach to window
 this.addEventListener(window, 'scroll', fn);
 ```


README.md
```diff
@@ -659,6 +659,9 @@ this.addEventListener('.someClass', 'scroll', fn);
 // Attach to a DOM node
 this.addEventListener(document.body, 'click', fn);
 
+// Attach to a component's root element
+this.addEventListener(this.element, 'scroll', fn);
```
scalvert: I'm not opposed to adding this generally, though it makes me wonder if we can do things a bit better for this specific case. 

I wonder if we should have an overload that assumes if no element is passed then we mean `this.element`.

```
this.addEventListener('scroll', fn);
```

Would this be useful?

steveszc: That makes a lot of sense to me.

scalvert: I'm going to bundle these changes up with part of a larger refactor I'm working on. I'll comment on this PR once I've made progress.

scalvert:
I've addressed this in #67. We now support the following API:

```js
addEventListener(this, 'click', fn);
```

This will use the component's element by default if no element is provided.

PR #66, opened 2018-02-08
Refactor to Weakmaps Part 2 - tasks to disposable

scalvert:
This PR represents the second part of porting lifeline to leveraging WeakMap instead of instance arrays.

TODO:

- [x] Convert `*Task` to use registerDisposable instead of instance array
  - [x] scheduleTask
  - [x] debounceTask
  - [x] throttleTask
  - [x] pollTask
- [x] Port tests from mixin to util
- [x] Make tests pass


scalvert:
@stefanpenner @rwjblue - this thing is good to go.

addon/debounce-task.js
```diff
@@ -0,0 +1,148 @@
...
+    let debounceNames = debounces && Object.keys(debounces);
+
+    if (!debounceNames || !debounceNames.length) {
+      return;
+    }
+
+    for (let i = 0; i < debounceNames.length; i++) {
+      let { cancelId } = debounces[debounceNames[i]];
+
+      cancelTask(cancelId);
```
rwjblue: Seems odd that this goes through `cancelTask` but the `cancelDebounce` method above uses `run.cancel(cancelId)` directly.  I would have expected them both to go through the same path...

addon/debounce-task.js
```diff
@@ -0,0 +1,148 @@
...
+  assert(
+    `Called \`obj.debounceTask('${name}', ...)\` where 'obj.${name}' is not a function.`,
+    typeof obj[name] === 'function'
+  );
+  assert(
+    `Called \`debounceTask\` on destroyed object: ${obj}.`,
+    !obj.isDestroyed
+  );
+
+  let pendingDebounces = getOrAllocate(
```
rwjblue: Does `getOrAllocate` actually provide value here? To my eyes it is more "magical" or "confusing" than the underlying implementation without it would be.

Consider:

```js
let pendingDebounces = getOrAllocate(
    registeredDebounces,
    obj,
    Object,
    getDebouncesDisposable
  );
```

versus:

```js
let pendingDebounces = registeredDebounces.get(obj);

if (!pendingDebounces) {
  pendingDebounces = {};
  registeredDebounces.set(obj, pendingDebounces);
  registerDisposable(obj, getDebouncesDisposable(pendingDebounces));
}
```

The second example is much clearer to me, and avoids passing the weakmap externally...

IMHO, we should remove the `getOrAllocate` utility...

addon/poll-task.js
```diff
@@ -0,0 +1,168 @@
...
+                                           which is run at the provided time specified
+                                           by timeout
+   @public
+   */
+export function pollTask(obj, taskOrName, token = getNextToken()) {
+  let next;
+  let task = getTask(obj, taskOrName, 'pollTask');
+  let tick = () => task.call(obj, next);
+
+  getOrAllocate(registeredPollers, obj, Array, getPollersDisposable).push(
```
rwjblue: same general feels RE getOrAllocate here...

addon/debounce-task.js
scalvert: I'm fine with explicitly calling `run.cancel` here to. Makes no diff to me.

addon/debounce-task.js
scalvert: I could be persuaded, though it _is_ duplication of the same logic.

addon/poll-task.js
scalvert: Doesn't a bit of DRY help those feels?

rwjblue: NO! 🙃 

IMHO, each abstraction should add value or it shouldn't exist. This specific one makes the actual logic _harder_ to understand...

addon/debounce-task.js
scalvert: I'm going to change it. I hate that signature.

addon/debounce-task.js
scalvert: Fixed.

addon/debounce-task.js
scalvert: Inlined for `debounce-task` and `poll-task`. I created a module scoped function for `run-task`, as the same logic was shared 3 times in that module.

addon/poll-task.js
scalvert: Fixed.

PR #68, opened 2018-03-06
Adding test support to ensure disposables have been run.

scalvert:
With the new functional API, users of the library run `registerDisposable` in order to queue a disposable to run. They're then required to ensure that `runDisposables` is called at some point during the destruction phase. Currently, there's no way to ensure that users correctly implement their objects in such a way as to guarantee that disposables are called. When not implemented, this library becomes useless.

As a way to help ensure disposables are called, this PR introduces a test helper which, when setup in your tests, will hook into QUnit's `afterEach` and `testDone` respectively and ensure you don't have any disposables that haven't run. 

- Adding refcounter for disposables to ensure they run. 
- Added tests. 
- Added test integration

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,0 +1,28 @@
+import QUnit from 'qunit';
+import { hasRunAllDisposables } from 'ember-lifeline';
+
+const FAILED_ASSERTION_MESSAGE =
+  'One or more objects registered disposables that were not correctly disposed of. Please ensure that objects correctly run their registered disposables by calling `runDisposables` in the `destroy` method of the object.';
+let setupTestDone = false;
+
+export default function setupEnsureLifelineDisposablesRun(hooks) {
+  if (!setupTestDone) {
+    QUnit.testDone(ensureDisposablesRunIfNoFailures);
+
+    setupTestDone = true;
+  }
+
+  hooks.afterEach(ensureDisposablesRun);
+}
+
+export function ensureDisposablesRunIfNoFailures(details) {
+  if (details.failed === 0) {
+    ensureDisposablesRun();
+  }
+}
+
+export function ensureDisposablesRun() {
+  if (!hasRunAllDisposables()) {
+    QUnit.assert.ok(false, FAILED_ASSERTION_MESSAGE);
+  }
+}
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,11 +1,13 @@
-import QUnit from 'qunit';
+import require from 'require';
 import { hasRunAllDisposables } from 'ember-lifeline';
+
+let QUnit = require('qunit').default;
 
 const FAILED_ASSERTION_MESSAGE =
   'One or more objects registered disposables that were not correctly disposed of. Please ensure that objects correctly run their registered disposables by calling `runDisposables` in the `destroy` method of the object.';
 let setupTestDone = false;
 
-export default function setupEnsureLifelineDisposablesRun(hooks) {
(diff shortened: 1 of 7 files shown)


addon-test-support/index.js
```diff
@@ -0,0 +1,28 @@
+import QUnit from 'qunit';
```
rwjblue: We can't assume `QUnit` is available here, can you swap instead to lazily importing?

addon-test-support/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+    setupTestDone = true;
+  }
+
+  hooks.afterEach(ensureDisposablesRun);
+}
+
+export function ensureDisposablesRunIfNoFailures(details) {
+  if (details.failed === 0) {
+    ensureDisposablesRun();
```
rwjblue: I think this needs to be a bit different. We can't call `QUnit.assert.ok` in a `testDone`.

Instead, I think it should be something like:

```js
if (!hasRunAllDisposables()) {
  throw new Error(FAILED_ASSERTION_MESSAGE);
}
```

Or something like that...

addon-test-support/index.js
```diff
@@ -0,0 +1,28 @@
+import QUnit from 'qunit';
+import { hasRunAllDisposables } from 'ember-lifeline';
+
+const FAILED_ASSERTION_MESSAGE =
+  'One or more objects registered disposables that were not correctly disposed of. Please ensure that objects correctly run their registered disposables by calling `runDisposables` in the `destroy` method of the object.';
+let setupTestDone = false;
+
+export default function setupEnsureLifelineDisposablesRun(hooks) {
```
rwjblue: `setupLifelineValidation`? 🚲 🏠

scalvert: What does validation mean to the end user?

addon-test-support/index.js
```diff
@@ -1,11 +1,13 @@
-import QUnit from 'qunit';
+import require from 'require';
 import { hasRunAllDisposables } from 'ember-lifeline';
 
+let QUnit = require('qunit').default;
```
rwjblue: this require should be done down inside of the `if (!setupTestDone) {` guard in `setupLifelineValidation` (which AFAICT is the only thing that is actually QUnit specific)

Also, if you update that to something like:

```js
if (!setupTestDone && require.has('qunit')) {
  let QUnit = require('qunit').default;
  QUnit.testDone(ensure....);
}
```

addon-test-support/index.js
```diff
@@ -0,0 +1,30 @@
+import require from 'require';
+import { hasRunAllDisposables } from 'ember-lifeline';
+
+let QUnit = require('qunit').default;
```
rwjblue: https://github.com/rwjblue/ember-lifeline/pull/68#discussion_r172581655 is still an issue here I believe

tests/unit/utils/disposable-test.js
```diff
@@ -2,26 +2,42 @@ import EmberObject from '@ember/object';
...
         runDisposables(this);
+
+        this._super(...arguments);
       },
-    });
+    }).create();
+
+    this.registeredDisposables = new Map();
+
+    _setRegisteredDisposables(this.registeredDisposables);
```
rwjblue: Can you add an `hooks.after(() => _setRegisteredDisposables(new WeakMap())` (which will run after all tests in this module have completed) to ensure we properly reset back to a WeakMap?

scalvert: Done. Also reset in `afterEach`

PR #71, opened 2018-03-18
Look for a string when asserting something is an element

### LINKED ISSUE #70: Unable to addEventListener to window in 3.0.x
This code no longer works after updating to v3.0.0

`this.addEventListener(window, 'mousemove', fn);` 

It triggers the assertion at https://github.com/rwjblue/ember-lifeline/blob/master/addon/dom-event-listeners.js#L147 as the `window` does not have a `nodeType` property.

jrjohnson:
nodeType does not exist on the window object so this assertion needs to
be relaxed a little in order to allow adding listeners to window. Fixes #70

I'm not sure this assertion needs to be relaxed *this much*, but this seemed cleaner than implementing a complicated duck type test since it will catch any common mistake given that we can't use `instanceof Element` https://github.com/rwjblue/ember-lifeline/pull/67#discussion_r171745877

### DIFF
--- a/addon/dom-event-listeners.js
+++ b/addon/dom-event-listeners.js
@@ -144,7 +144,13 @@ export function removeEventListener(
 
 function assertArguments(element, eventName, callback) {
   assert('Must provide a DOM element', !!element);
-  assert('Must provide an element (not a DOM selector)', !!element.nodeType);
+  assert(
+    'Must provide an element (not a DOM selector)',
+    typeof element !== 'string' &&
+      typeof element !== 'function' &&
+      element !== null &&
+      element !== undefined
+  );
   assert(
     'Must provide an eventName that specifies the event type',
     !!eventName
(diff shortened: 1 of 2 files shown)


addon/dom-event-listeners.js
```diff
@@ -144,7 +144,10 @@ export function removeEventListener(
 
 function assertArguments(element, eventName, callback) {
   assert('Must provide a DOM element', !!element);
-  assert('Must provide an element (not a DOM selector)', !!element.nodeType);
+  assert(
+    'Must provide an element (not a DOM selector)',
+    typeof element !== 'string'
```
rwjblue: Should also disallow `undefined` and `null` at least.

rwjblue: (And functions to ensure the signature is correct)

jrjohnson:
@rwjblue no problem... done.

PR #73, opened 2018-03-21
Exit cancelDebounce if task is no longer pending

magistrula:
Fixes the following issue:
1. Trigger a debounced task and wait for it to finish running.
2. Perform some other action in the UI that cancels any debounced tasks that might be running.
Expected Result: Nothing happens, since there is nothing to cancel
Actual Result: `TypeError: Cannot read property 'cancelId' of undefined`

### DIFF
--- a/addon/debounce-task.js
+++ b/addon/debounce-task.js
@@ -123,7 +123,7 @@ export function debounceTask(obj, name, ...debounceArgs) {
 export function cancelDebounce(obj, name) {
   let pendingDebounces = registeredDebounces.get(obj);
 
-  if (!pendingDebounces || !pendingDebounces[name]) {
+  if (pendingDebounces === undefined || pendingDebounces[name] === undefined) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


addon/debounce-task.js
```diff
@@ -123,7 +123,7 @@ export function debounceTask(obj, name, ...debounceArgs) {
 export function cancelDebounce(obj, name) {
   let pendingDebounces = registeredDebounces.get(obj);
 
-  if (pendingDebounces === undefined) {
+  if (!pendingDebounces || !pendingDebounces[name]) {
```
rwjblue: The change seems good to me! Would you mind refactoring to something like:

```js
if (pendingDebounces === undefined || pendingDebounces[name] === undefined) {
```

I'd like to (in general) only guard against the specific values we expect (the only valid values are `undefined` and the actual expected object)...

magistrula: 👍Done. I was curious about that; thanks for explaining the rationale behind checking specifically for `undefined`.

rwjblue:
Thank you!

PR #75, opened 2018-03-23
Fix pollTaskFor

### LINKED ISSUE #39: Maybe pollTaskFor should wait()?
My tests fail if I don't `wait()` before and after `pollTaskFor`:
```js
await wait();
await pollTaskFor('label');
await wait();
```

The tests in this repo also seem to doing that. Any reason it can't be rolled into `pollTaskFor`?

scalvert:
Fixes #39

### DIFF
--- a/addon/poll-task.js
+++ b/addon/poll-task.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import { run } from '@ember/runloop';
 import { assert } from '@ember/debug';
+import { settled } from '@ember/test-helpers';
 import getTask from './utils/get-task';
 import { registerDisposable } from './utils/disposable';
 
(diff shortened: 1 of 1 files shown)


addon/poll-task.js
```diff
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import { run } from '@ember/runloop';
 import { assert } from '@ember/debug';
+import { settled } from '@ember/test-helpers';
```
rwjblue: This will fail in non-test code. We should refactor things a bit so that the test related code actually comes from addon-test-support instead of addon/

scalvert: Agree, which is why I pinged you yesterday about this :)

scalvert:
Gonna close this in favor of refactoring `pollTaskFor` out of its current location.

PR #93, opened 2018-04-09
clean up cancelId before running task to avoid memory leak

### LINKED ISSUE #92: memory leak in scheduleTask
weak map is not being cleaned up.

`getTimersDisposable` leaves an entry in the `registeredTimers` WeakMap and individual `cancelIds` are left after tasks have completed

BryanCrotaz:
Fixes #92

### DIFF
--- a/addon/run-task.js
+++ b/addon/run-task.js
@@ -13,7 +13,18 @@ const { WeakMap } = Ember;
  * @private
  *
  */
-const registeredTimers = new WeakMap();
+let registeredTimers = new WeakMap();
+
+/**
+ * Test use only. Allows for swapping out the WeakMap to a Map, giving
+ * us the ability to detect whether the timers set is empty.
+ *
+ * @private
+ * @param {*} mapForTesting A map used to ensure correctness when testing.
+ */
+export function _setRegisteredTimers(mapForTesting) {
+  registeredTimers = mapForTesting;
+}
 
 /**
    Registers and runs the provided task function for the provided object at the specified
--- a/addon/run-task.js
+++ b/addon/run-task.js
@@ -115,7 +115,7 @@ export function scheduleTask(obj, queueName, taskOrName, ...args) {
 
   let task = getTask(obj, taskOrName, 'scheduleTask');
   let timers = getTimers(obj);
-  var cancelId;
+  let cancelId;
   let taskWrapper = (...taskArgs) => {
     // clean up
     let index = timers.indexOf(cancelId);
@@ -123,7 +123,7 @@ export function scheduleTask(obj, queueName, taskOrName, ...args) {
       timers.splice(index, 1);
     }
     task.call(obj, ...taskArgs);
-  }
+  };
   cancelId = run.schedule(queueName, obj, taskWrapper, ...args);
 
   timers.push(cancelId);
@@ -162,7 +162,7 @@ export function scheduleTask(obj, queueName, taskOrName, ...args) {
    @param { Number } [timeout] the time in the future to run the task
    @public
    */
-export function throttleTask(obj, name, timeout = 0) {
+export function throttleTask(obj, taskOrName, timeout = 0) {
   assert(
     `Called \`throttleTask\` without a string as the first argument on ${obj}.`,
     typeof name === 'string'
@@ -177,7 +177,7 @@ export function throttleTask(obj, name, timeout = 0) {
   );
 
   let timers = getTimers(obj);
-  var cancelId;
+  let cancelId;
   let task = getTask(obj, taskOrName, 'throttleTask');
   let taskWrapper = (...taskArgs) => {
     let index = timers.indexOf(cancelId);
@@ -185,7 +185,7 @@ export function throttleTask(obj, name, timeout = 0) {
       timers.splice(index, 1);
     }
     task.call(obj, ...taskArgs);
-  }
+  };
   cancelId = run.throttle(obj, taskWrapper, timeout);
 
   timers.push(cancelId);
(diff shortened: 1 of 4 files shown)


BryanCrotaz:
Not sure how to go about testing this - the `registeredTimers` are private to the run-task file - how would a test get at them?

BryanCrotaz:
I don't think `cancelTask` has enough information to clean up the ids array - needs looking at.

BryanCrotaz:
Also improved `throttleTask` to accept a function or a method name

scalvert:
Much appreciate the PR, @BryanCrotaz! I believe our view was that the WeakMap wouldn't necessarily represent an egregious leak due to the objects not being long lived. It does, however, seem like a bit of a smell to have cancelIds kicking around past their due date...

I'll take a look at this PR and add comments as needed. 

WRT the tests, and validating that we've cleared the WeakMap, I've done something similar where I actually optionally set the WeakMap via a hook that's used for testing, which in turn allows me to reference the map from testland.

```javascript
let _registeredTimers = new WeakMap();
export function setRegisteredTimers(map) {
  _registeredTimers = map;
}
```

This allows tests to pass in the map and retain the reference, while still providing the same functionality to the util. Would this work for you?

BryanCrotaz:
I'm using it to bring websocket messages into the run loop using scheduleTask, so I'm building up several MB of cancelIds without this fix, causing the browser to die after less than 24 hours.

BryanCrotaz:
```js
let _registeredTimers = new WeakMap();
export function _setRegisteredTimers(map) {
  _registeredTimers = map;
}
```

Sneaky - I like it.

stefanpenner:
@BryanCrotaz thanks for catching this, yes we should fix.

addon/run-task.js
```diff
@@ -114,8 +114,17 @@ export function scheduleTask(obj, queueName, taskOrName, ...args) {
   );
 
   let task = getTask(obj, taskOrName, 'scheduleTask');
-  let cancelId = run.schedule(queueName, obj, task, ...args);
   let timers = getTimers(obj);
+  let cancelId;
+  let taskWrapper = (...taskArgs) => {
+    // clean up
+    let index = timers.indexOf(cancelId);
```
scalvert: It seems like there may be a bit of duplication between `runTask` and this. Could we just toss that into a function?

addon/run-task.js
```diff
@@ -153,22 +162,27 @@ export function scheduleTask(obj, queueName, taskOrName, ...args) {
...
-export function throttleTask(obj, name, timeout = 0) {
-  assert(
-    `Called \`throttleTask\` without a string as the first argument on ${obj}.`,
-    typeof name === 'string'
-  );
-  assert(
-    `Called \`throttleTask('${name}', ${timeout})\` where '${name}' is not a function.`,
-    typeof obj[name] === 'function'
-  );
+export function throttleTask(obj, taskOrName, timeout = 0) {
```
scalvert: I'd prefer to keep these changes separate, as they're not really related to this PR per se. Is it OK if we move them out?

addon/run-task.js
```diff
@@ -207,12 +221,18 @@ export function throttleTask(obj, name, timeout = 0) {
    */
 export function cancelTask(cancelId) {
```
scalvert: So this seems to imply we need to change the signature of this function to take an object in addition to the cancelId. Correct?

tests/unit/run-task-test.js
```diff
@@ -164,21 +164,35 @@ module('ember-lifeline/run-task', function(hooks) {
     assert.notOk(hasRun, 'callback should have been canceled previously');
   });
 
-  test('throttleTask triggers an assertion when a string is not the first argument', function(assert) {
+  test('throttleTask does not trigger an assertion when a function is the first argument', function(assert) {
```
scalvert: Again, I'd pull these out so as to not conflate the changes.

scalvert:
@BryanCrotaz thanks for working on this. I make a few changes to the PR, mainly reverting the changes to `throttleTask`. I also changed the `timers` array to a `Set`, which allows for us to use a cleaner syntax when managing removing the `cancelIds`. 

The reason we want to leave `throttleTask` as is is that we want to force people to use the string name API for `throttle`, as it guarantees a stable function when used in conjunction with throttling. If we allow people to pass in functions, there's a potential that they don't use a stable function, essentially removing the utility of throttling in the first place.

scalvert:
Can you sanity check, @rwjblue?

BryanCrotaz:
Needs a deprecation warning in cancelTask

rwjblue:
> Needs a deprecation warning in cancelTask

Nothing is intended to be deprecated just yet...

BryanCrotaz:
@rwjblue then a warning that you will get a memory leak if you don't use the `obj` argument?
The TODO comment in `cancelTask` implies that the old form is deprecated.

rwjblue:
I think in most cases it doesn't matter (leaking a single integer over time is not really an issue), and in the cases where it does folks can update to the other format.

I'm personally not opposed to actually deprecating `cancelTask(cancelId)`, but I'd rather not incorporate that deprecation along with this bugfix. Instead, it should be evaluated independently as a separate PR.  Not sure how @scalvert feels though...

BryanCrotaz:
Any memory leak is an issue in a long lived app, especially on memory constrained devices (mine has 500MB of total system RAM). I'll submit a new PR.

scalvert:
So I'm inclined to think adding the deprecation warning should be added, but definitely in a separate PR. I'm totally open to merging it if you submit it, @BryanCrotaz. Also, thanks for bringing this to our attention!

BryanCrotaz:
@scalvert deprecate until Ember 4.0?

BryanCrotaz:
See #96 and #97

PR #96, opened 2018-04-17
add deprecation warning and documentation for new form of cancelTask

BryanCrotaz:


addon/run-task.js
```diff
@@ -4,7 +4,7 @@ import { assert } from '@ember/debug';
 import getTask from './utils/get-task';
 import { registerDisposable } from './utils/disposable';
 
-const { WeakMap } = Ember;
+const { WeakMap, deprecate } = Ember;
```
scalvert: Can we import `deprecate` from `@ember/debug`? We're already importing `assert`.

addon/run-task.js
```diff
@@ -215,25 +215,32 @@ export function throttleTask(obj, name, timeout = 0) {
    ```
 
    @method cancelTask
+   @param { Object } obj the entwined object that was provided with the original runTask or scheduleTask call
```
scalvert: Can we use the word `entangled` here instead? This would match the wording used in the README.

addon/run-task.js
```diff
@@ -215,25 +215,32 @@ export function throttleTask(obj, name, timeout = 0) {
    ```
 
    @method cancelTask
+   @param { Object } obj the entangled object that was provided with the original runTask or scheduleTask call
```
scalvert: `the entangled object that was provided with the original *Task call`

PR #97, opened 2018-04-17
remove memory leak in pollTask

BryanCrotaz:


addon/poll-task.js
```diff
@@ -158,15 +158,32 @@ export function pollTask(obj, taskOrName, token = getNextToken()) {
...
+      {
+        id: 'ember-lifeline-cancel-poll-without-object',
+        until: '4.0.0',
+      }
+    );
+    token = obj;
+    obj = null;
+  }
   delete queuedPollTasks[token];
+  if (obj) {
```
rwjblue: I'd prefer this to become an `else` to the `if (token === undefined)` conditional above

addon/poll-task.js
```diff
@@ -158,15 +158,32 @@ export function pollTask(obj, taskOrName, token = getNextToken()) {
...
+  }
 }
 
 function getPollersDisposable(pollers) {
   return function() {
-    for (let i = 0; i < pollers.length; i++) {
-      delete queuedPollTasks[pollers[i]];
-    }
+    pollers.forEach(token => {
+      cancelPoll(token);
```
rwjblue: Doesn't this trigger the deprecation during destruction of every object that uses polling? In order to add the deprecation, we have to ensure that our own internal usages do _not_ trigger it...

@scalvert - We should bring in some additional test helper utilities to ensure none of our tests trigger deprecations ([example here](https://github.com/emberjs/ember-test-helpers/blob/7f9a3d082958ba993c6db863886d370d73fd98ef/tests/test-helper.js#L36-L102))...

addon/poll-task.js
```diff
@@ -158,15 +158,32 @@ export function pollTask(obj, taskOrName, token = getNextToken()) {
    @param { String } token the token for the pollTask to be cleared
    @public
    */
-export function cancelPoll(token) {
+export function cancelPoll(obj, token) {
+  // TODO: remove this older API on next major version
```
rwjblue: I don't think we need the TODO since this also adds the deprecations (all deprecations are TODO's for a major version bump)...

addon/poll-task.js
```diff
@@ -4,7 +4,7 @@ import { assert } from '@ember/debug';
 import getTask from './utils/get-task';
 import { registerDisposable } from './utils/disposable';
 
-const { WeakMap } = Ember;
+const { WeakMap, deprecate } = Ember;
```
scalvert: Can we import `deprecate` from `@ember/debug`? We already do this for `assert`.

scalvert:
Thanks, @BryanCrotaz! Just one small comment.

scalvert:
On second thought we don't explicitly cancel `pollTask` when invoked, as it relies on subsequent `runTask` calls. This looks good to me.

addon/poll-task.js
scalvert: Agree.

BryanCrotaz:
Thanks

scalvert:
@BryanCrotaz thanks for opening this!

PR #101, opened 2018-05-25
Converts codebase to TypeScript

scalvert:


addon/debounce-task.ts
```diff
@@ -1,9 +1,13 @@
-import Ember from 'ember';
+import EmberObject from '@ember/object';
 import { run } from '@ember/runloop';
```
krisselden: don't use run as a namespace, you can import the debounce, cancel etc

scalvert: Fixed!

PR #123, opened 2018-06-12
Readme list formatting

lolmaus:
These lines get collapsed into a single line due to how Mardown deals with line breaks.

A proper way to format them is a list. Alternatively, double spaces can be added to the end of each line in order to prevent collapsing.


lolmaus:
Before:

> image

***

After: 

> image

rwjblue:
Awesome, thank you!

PR #141, opened 2018-07-19
Rerunning `runDisposables`

nathanhammond:
Not all use cases for `ember-lifeline` occur on `destroy`. For example, `Ember.Route#activate`/`Ember.Route#deactivate` may be a place where you want this behavior.

This is a fix for DOM listener multiple teardowns.

### DIFF
--- a/addon/dom-event-listeners.js
+++ b/addon/dom-event-listeners.js
@@ -88,7 +88,17 @@ export function addEventListener(obj, element, eventName, callback, options) {
   assertArguments(element, eventName, callback);
 
   let _callback = run.bind(obj, callback);
-  let listeners = getEventListeners(obj);
+  let listeners = eventListeners.get(obj);
+
+  if (listeners === undefined) {
+    listeners = [];
+    eventListeners.set(obj, listeners);
+  }
+
+  // Register a disposable every time we go from zero to one.
+  if (listeners.length === 0) {
+    registerDisposable(obj, getEventListenersDisposable(listeners));
+  }
 
   if (!PASSIVE_SUPPORTED) {
     options = undefined;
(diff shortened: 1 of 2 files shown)


addon/dom-event-listeners.js
```diff
@@ -170,7 +170,7 @@ function getEventListenersDisposable(listeners) {
 
         element.removeEventListener(eventName, callback, options);
       }
-      listeners.length = 0;
+      eventListeners.set(obj, undefined);
```
rwjblue: Why set to `undefined` (vs remove completely)?

nathanhammond: No particular reason, I got into this local maxima because I looked at the previous version which erroneously set `listeners = undefined` which just changed the reference but didn't update the thing which the `WeakMap` pointed to.

I'll take a swing at a real solution which actually takes advantage of the perf optimizations that exist here. (`WeakMap` target is an object with a `listeners` array.)

scalvert:
Thanks, Nathan. Just to be clear, from your description do you mean that it's possible that you would want `runDisposables` to be invoked _both_ in `destroy` and additionally elsewhere, such as `activate/deactivate`?

It looks as though you may have uncovered an odd intersection between the WeakMap in `dom-event-handlers` and the running of the disposable. If we feel the need to support being able to run `runDisposables` multiple times, I think your change makes sense.

rwjblue:
Ya, I think we should support running disposables multiple times (activate/deactivate is a good motivating example). We should probably beef up the tests for all of the utility functions confirming that they all work properly with this constraint.

nathanhammond:
I've got a different fix for this bug in place that doesn't reallocate `listeners` if it already exists. Same number of conditional checks, plus one more to catch the condition that triggered the bug.

I've not yet reviewed the other cases to see if they work, but I'll take a look at my app in a second.

nathanhammond:
@scalvert Confirm, I want `runDisposables` to be triggered in any number of locations for long-lived objects such as routes on `activate`/`deactivate`. (My actual, real-world use case that brought me here today.) And also on `destroy`, because testing.

nathanhammond:
Might I bother y'all for a point release?

scalvert:
Done. v3.0.5 published 🎉

PR #148, opened 2018-07-25
Add optional `map` property to expose the `registeredDisposables` map to tests.

nathanhammond:
This is a point of discussion, though I figured it would be easier to discuss with a concrete proposal.

The idea here is to be able to have insight into the registered disposables in testing. It consequentially exposes a lot more `ember-lifeline` internals to possible test-user-land code. That's my goal, but it of course has tradeoffs in terms of either:
- Implied public API which now has to remain stable.
- Exposed non-public API which is unstable that people may depend on nonetheless.

It's a power-user feature, so documenting it in the `README` may not be something we want to do. Obscuring its existence may be good enough to say, "hey, you're on your own here."

I'm not tremendously attached to this proposal, and strong opinions in opposition to this (if they exist) should carry the day.

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -5,8 +5,23 @@ const FAILED_ASSERTION_MESSAGE =
   'One or more objects registered disposables that were not correctly disposed of. Please ensure that objects correctly run their registered disposables by calling `runDisposables` in the `destroy` method of the object.';
 let setupTestDone = false;
 
-export default function setupLifelineValidation(hooks) {
-  let registeredDisposables = new Map();
+/**
+ * Test helper to assert that all async work is disposed of.
+ *
+ * @method setupLifelineValidation
+ * @param { QUnit.hooks } hooks Qunit's hooks object.
+ * @param { Map } options.map Optional map object to use for external reference.
+ * @public
+ */
+export default function setupLifelineValidation(hooks, options) {
+  let registeredDisposables;
+
+  if (options && options.map instanceof Map) {
+    registeredDisposables = options.map;
+  } else {
+    registeredDisposables = new Map();
+  }
+
   hooks.beforeEach(function() {
     _setRegisteredDisposables(registeredDisposables);
   });
@@ -31,8 +46,7 @@ export default function setupLifelineValidation(hooks) {
 
       assert.deepEqual(retainedObjects, [], FAILED_ASSERTION_MESSAGE);
     } finally {
-      registeredDisposables = new WeakMap();
-      _setRegisteredDisposables(registeredDisposables);
+      _setRegisteredDisposables(new WeakMap());
     }
   });
 }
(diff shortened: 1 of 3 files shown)


addon-test-support/index.js
```diff
@@ -5,8 +5,23 @@ const FAILED_ASSERTION_MESSAGE =
...
+ *
+ * @method setupLifelineValidation
+ * @param { QUnit.hooks } hooks Qunit's hooks object.
+ * @param { Map } options.map Optional map object to use for external reference.
+ * @public
+ */
+export default function setupLifelineValidation(hooks, options) {
+  let registeredDisposables;
+
+  if (options && options.map instanceof Map) {
```
rwjblue: I don't think we need to check what type of object this is, if you give us an object that duck-types correctly we should just use it...

addon-test-support/index.js
```diff
@@ -5,8 +5,23 @@ const FAILED_ASSERTION_MESSAGE =
...
 let setupTestDone = false;
 
-export default function setupLifelineValidation(hooks) {
-  let registeredDisposables = new Map();
+/**
+ * Test helper to assert that all async work is disposed of.
+ *
+ * @method setupLifelineValidation
+ * @param { QUnit.hooks } hooks Qunit's hooks object.
+ * @param { Map } options.map Optional map object to use for external reference.
```
rwjblue: Should be `options._map` (underscored to make it clear that this is definitely private API)...

addon-test-support/index.js
```diff
@@ -31,8 +46,7 @@ export default function setupLifelineValidation(hooks) {
 
       assert.deepEqual(retainedObjects, [], FAILED_ASSERTION_MESSAGE);
     } finally {
-      registeredDisposables = new WeakMap();
```
rwjblue: This should be brought back (or we will possibly leak the provided map)...

tests/unit/setup-lifeline-validation-options-test.js
```diff
@@ -0,0 +1,32 @@
...
+  let disposableMap = new Map();
+  setupLifelineValidation(hooks, { map: disposableMap });
+  setupTest(hooks);
+
+  hooks.beforeEach(function(assert) {
+    let serviceName = 'service:under-test';
+    this.owner.register(serviceName, Service.extend());
+    let factory = this.owner.factoryFor
+      ? this.owner.factoryFor(serviceName)
+      : this.owner._lookupFactory(serviceName);
```
rwjblue: 😩 do we still support Ember < 2.12?!?!?!

scalvert:
Can you enumerate the use cases that would precipitate you needing this feature? I can't think of why you'd want to have this capability.

nathanhammond:
So, I threw this out here, but after I did it, I found a better way to architect my own code for testing that made it superfluous. Something like this _may_ be required to get better confidence about what's actually happening (I'm missing a a couple assertions I would like to make in a unit test), but it's not entirely necessary to peek that far into the black box. I was able to effectively test those code paths at a higher level.

I'm closing this, I don't think it's valuable enough to make it worth the potential brittleness cost.

scalvert:
Nice. Glad you were able to work out what you needed. Feel free to reopen if you feel like there's a valid use case!

PR #160, opened 2018-08-05
Bugfix: checking for pollers to be defined before calling cancelPoll

scalvert:
Calling `cancelPoll` before calling `pollTask` results in `pollers` being undefined, as the set is only initialized via `pollTask`. This fix adds a check to ensure `pollers` is not undefined before calling delete on the set.

### DIFF
--- a/tests/unit/poll-task-test.js
+++ b/tests/unit/poll-task-test.js
@@ -17,6 +17,8 @@ module('ember-lifeline/poll-task', function(hooks) {
   hooks.beforeEach(function() {
     this.BaseObject = EmberObject.extend({
       destroy() {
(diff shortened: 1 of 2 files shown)


tests/unit/poll-task-test.js
```diff
@@ -17,6 +17,8 @@ module('ember-lifeline/poll-task', function(hooks) {
   hooks.beforeEach(function() {
     this.BaseObject = EmberObject.extend({
       destroy() {
+        this._super(...arguments);
```
rwjblue: Generally prefer super _after_ custom stuff when in destroy (since it is the inverse of unit/constructor). Doesn’t really matter though....

Issue #162, opened 2018-08-05
addEventListener cannot bind events directly to window or svg elements in 3.0.6


pgrippi:
The [findElement](https://github.com/ember-lifeline/ember-lifeline/blob/master/addon/mixins/dom.ts#L141) method was updated in 3.0.6 to use this check to see if an element was passed into `addEventListener`:

```
selector instanceof HTMLElement && (selector.nodeType || selector instanceof Window)
```

Both window and svg elements fail to pass the first instanceof check

rwjblue:
Code from prior version:

https://github.com/ember-lifeline/ember-lifeline/blob/89a1bb823874ddd5f626cb05a66cd8600ef6d5e9/addon/mixins/dom.ts#L140-L147

rwjblue:
Once we fix, we should add tests for both SVG and window to ensure we do not regress again...

scalvert:
Resolved in v3.0.7 🎉

PR #164, opened 2018-08-06
Bugfix: fixes issue with TS upgrade introducing issue with adding events to window/svg

scalvert:
Addresses #162

### DIFF
--- a/addon/mixins/dom.ts
+++ b/addon/mixins/dom.ts
@@ -144,12 +144,11 @@ function findElement(
 ): HTMLElement {
   let selectorType: string = typeof selector;
   let element;
-
+debugger;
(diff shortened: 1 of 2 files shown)


scalvert:
cc: @pgrippi

addon/mixins/dom.ts
```diff
@@ -144,12 +144,11 @@ function findElement(
 ): HTMLElement {
   let selectorType: string = typeof selector;
   let element;
-
+debugger;
```
rwjblue: 🔪

scalvert: oh good lord...

scalvert:
They were. Both were failing.

PR #195, opened 2018-09-14
Assign registeredDisposables to Map beforeEach test

### LINKED ISSUE #194: "registeredDisposables.keys is not a function" when two tests in one module
When I run these two tests I get an error on the second test:

```
afterEach failed on visiting /posts : registeredDisposables.keys is not a function
```

```js
import { module, test } from 'qunit';
import { setupApplicationTest } from 'ember-qunit';
import { visit, currentURL } from '@ember/test-helpers';
import setupLifelineValidation from 'ember-lifeline/test-support';

module('Acceptance | clean up', hooks => {
  setupLifelineValidation(hooks);
  setupApplicationTest(hooks);

  test('visiting /posts', async assert => {
    await visit('/posts');
    assert.equal(currentURL(), '/posts');
  });

  test('visiting /posts', async assert => {
    await visit('/posts');
    assert.equal(currentURL(), '/posts');
  });
});
```

It looks like this is because in the afterEach hook, `registeredDisposables` is assigned to a `WeakMap` in the same closure as the `beforeEach`, causing the cleanup step to blow up.

The solution is probably a one-line assignment to `registeredDisposables = new Map();` in `hooks.beforeEach` if I understand the problem and intention of this hook correctly .

mehulkar:
Closes #194.

### DIFF
--- a/tests/acceptance/cleanup-test.js
+++ b/tests/acceptance/cleanup-test.js
@@ -1,0 +1,19 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { visit, currentURL } from '@ember/test-helpers';
+import setupLifelineValidation from 'ember-lifeline/test-support';
+
+module('Acceptance | cleaning up disposables', function(hooks) {
+  setupLifelineValidation(hooks);
+  setupApplicationTest(hooks);
+
+  test('visiting /', async function(assert) {
+    await visit('/');
+    assert.equal(currentURL(), '/');
+  });
+
+  test('visiting /', async function(assert) {
+    await visit('/');
+    assert.equal(currentURL(), '/');
+  });
+});
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -31,8 +31,7 @@ export default function setupLifelineValidation(hooks) {
 
       assert.deepEqual(retainedObjects, [], FAILED_ASSERTION_MESSAGE);
     } finally {
-      registeredDisposables = new WeakMap();
-      _setRegisteredDisposables(registeredDisposables);
+      _setRegisteredDisposables(new WeakMap());
     }
   });
 }


mehulkar:
@scalvert @rwjblue @nathanhammond could someone take a look and see if I'm misdiagnosing the problem?

mehulkar:
yep your fix works. updated and added failing test

tests/acceptance/cleanup-test.js
```diff
@@ -0,0 +1,19 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { visit, currentURL } from '@ember/test-helpers';
+import setupLifelineValidation from 'ember-lifeline/test-support';
+
+module('Acceptance | clean up', hooks => {
```
scalvert: Can we change this from arrow to function? I know it's a bit nit picky, but it's inline with the rest of our tests. Also, it adheres to this rule -> https://github.com/platinumazure/eslint-plugin-qunit/blob/master/lib/rules/no-arrow-tests.js, which I happen to agree with.

scalvert:
Thanks for catching this, @mehulkar. I just left one comment but otherwise it looks good.

tests/acceptance/cleanup-test.js
mehulkar: i'm all about the linting 👍. do you want me to add eslint to the project with that rule too? Not sure if addons are supposed to have their own rules.

mehulkar:
@scalvert updated, let me know if you want me to install that plugin and add that rule as well

tests/acceptance/cleanup-test.js
```diff
@@ -0,0 +1,19 @@
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { visit, currentURL } from '@ember/test-helpers';
+import setupLifelineValidation from 'ember-lifeline/test-support';
+
+module('Acceptance | cleaning up disposables', function(hooks) {
+  setupLifelineValidation(hooks);
+  setupApplicationTest(hooks);
+
+  test('visiting /', async assert => {
```
scalvert: Can we also change these to functions vs. arrow functions?

tests/acceptance/cleanup-test.js
```diff
@@ -0,0 +1,19 @@
...
+    await visit('/');
+    assert.equal(currentURL(), '/');
+  });
+
+  test('visiting /', async assert => {
```
scalvert: Can we also change these to functions vs. arrow functions?

scalvert:
A couple more places to update. Sorry, I assumed you'd change them all so only called out the one.

I think we can add that eslint plugin later in a separate PR, for sure.

scalvert:
Thanks!

mehulkar:
awesome, can we get a new patch release also?

PR #248, opened 2018-10-22
feat(debounceTask, throttleTask): accept function

buschtoens:
While writing tests for [`ember-lifeline-decorators`](https://github.com/buschtoens/ember-lifeline-decorators#debounce), I discovered (much to my horror! 😱), that my beloved `ember-lifeline` does not accept functions for `debounceTask` and `throttleTask`.

This PR removes this grievance. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Therefore, *`scheduleTask` is prohibited from scheduling work on the `afterRende
 
 ### `debounceTask`
 
-**tl;dr Call `debounceTask(obj, methodName, args*, wait, immediate)` on any object to debounce work.**
+**tl;dr Call `debounceTask(obj, methodNameOrFunction, args*, wait, immediate)` on any object to debounce work.**
 
 Debouncing is a common async pattern often used to manage user input. When a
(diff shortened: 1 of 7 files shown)


buschtoens:
Still very much interested in this PR. Can we merge it, if I rebase it?

rwjblue:
I think the decorators could work around this by expanding to _two_ functions...

buschtoens:
Hm. I understand your reasoning. Thanks for explaining.

Adding another "invisible" method to the class seems like a hack to me. Do you think prefixing it with a guid and setting it to non-enumerable is reasonably safe?

rwjblue:
> Do you think prefixing it with a guid and setting it to non-enumerable is reasonably safe?

Yes, probably.  You could also experiment with using a `Symbol()` (which would make it non-enumerable also) that is only known/available inside the decorator closure.

buschtoens:
Thought about that too, but is IE11 supported sufficiently via the polyfills?
Nevermind, I'll find out. Thanks! 😊

PR #249, opened 2018-10-22
fix(addEventListener): broaded type to accept EventTarget

buschtoens:
`addEventListener` and friends actually accept any object that implements the [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) interface.


scalvert:
Thanks for the PR, @buschtoens! Would you mind writing some tests so I can get an idea as to what you're trying to support here (outside of elements)?

buschtoens:
I have not changed anything functionality-wise. I only fixed the types, because not only `HTMLElement`, which implements `EventTarget`, is supported, but actually any object that implements that interface is. 

If you like, I can add a test asserting that any object implementing the `EventTarget` interface is a valid target.

Issue #288, opened 2018-12-11
Loosen type from EmberObject to interface


scalvert:
Currently, most of the functions provided require an `EmberObject` to be provided as a way to key the WeakMaps. We should loosen this type to be an interface that conforms to a specific shape of what we expect. This will help the move to using native classes.

rwjblue:
I don't fully understand what you mean, are you just talking about the typings?

scalvert:
Yes. Currently the type is EmberObject. I think it should be more loosely defined, even if just in name.

rwjblue:
Gotcha. Yeah, I think defining the specific interface that we expect is important. _Especially_ as we approach custom components (ala sparkles-component / glimmer components) which vary quite a bit from Ember.Object.

scalvert:
Agree.

PR #289, opened 2018-12-13
Converts EmberObject types to IDestroyable

scalvert:
This PR provides looser typing for lifeline functions.

### DIFF
--- a/addon/interfaces/index.ts
+++ b/addon/interfaces/index.ts
@@ -5,4 +5,9 @@ export interface IMap<K extends object, V> {
   set(key: K, value: V): this;
 }
 
+export interface IDisposable {
+  isDestroyed: boolean;
+  isDestroying: boolean;
+}
+
 export type TaskOrName = Function | string;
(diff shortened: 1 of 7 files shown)


scalvert:
Fixes #288

addon/interfaces/index.ts
```diff
@@ -5,4 +5,9 @@ export interface IMap<K extends object, V> {
   set(key: K, value: V): this;
 }
 
+export interface IDisposable {
```
rwjblue: Sorry for bikeshedding here, but I think the name doesn't quite work. Specifically, we _already_ use this terminology ("disposable") in ember-lifeline for a very different thing. In [registerDisposable](https://github.com/ember-lifeline/ember-lifeline#registerdisposable) we are calling the callback function to run a "disposable" (not the object argument itself).

Maybe we could use `IDestroyable` here (which matches the boolean property names better anyways)?

scalvert: I thought of this, but opted for `IDisposable` as I feel like it implies that it's able to be disposed. I can get the argument that the disposable is _actually_ the function getting called, not the object. The object is something that can be destroyed.

I think I've just argued the case for myself to change it...

PR #302, opened 2018-12-22
Adding documentation generation for API

scalvert:
Adding documentation.js to auto-generate API docs from our jsdoc comments.

[Rendered](https://github.com/ember-lifeline/ember-lifeline/blob/87b90cf63cedf5e4f20357b355bfa9588d11c6da/API.md)


scalvert:
I'd like to, but I see no option to omit the params from toc. I see that you've done that in @ember/test-helpers, but I can't see what magic incantation you used. Any thoughts?

scalvert:
So there's no option to suppress the Parameters in the API docs. I think we should just merge this. We can figure out if we _really_ want them suppressed later. Sound OK?

rwjblue:
Seems fine to me

PR #335, opened 2019-02-05
handle target that is already destroyed in scheduleTask

BryanCrotaz:
… and return dummy task instead of asserting

### DIFF
--- a/addon/run-task.ts
+++ b/addon/run-task.ts
@@ -137,18 +137,20 @@ export function scheduleTask(
     `Called \`scheduleTask\` while trying to schedule to the \`afterRender\` queue on ${destroyable}.`,
     queueName !== 'afterRender'
   );
-  assert(
-    `Called \`scheduleTask\` on destroyed object: ${destroyable}.`,
-    !destroyable.isDestroyed
-  );
 
   let task: Function = getTask(destroyable, taskOrName, 'scheduleTask');
   let timers: Set<EmberRunTimer> = getTimers(destroyable);
   let cancelId: EmberRunTimer;
-  let taskWrapper: Function = (...taskArgs: any[]) => {
-    timers.delete(cancelId);
-    task.call(destroyable, ...taskArgs);
-  };
+  let taskWrapper: Function;
+  if (destroyable.isDestroyed) {
+    // dummy task so that we return something consistent even if target is destroyed
+    taskWrapper = () => {};
+  } else {
+    taskWrapper = (...taskArgs: any[]) => {
+      timers.delete(cancelId);
+      task.call(destroyable, ...taskArgs);
+    };
+  }
   cancelId = schedule(queueName, destroyable as any, taskWrapper, ...args);
 
   timers.add(cancelId);


BryanCrotaz:
Because the point of these functions is to stop the developer having to check isDestroying and isDestroyed. 

For example a DOM event which sets state on the object. scheduleTask(this, ... - these can often fire after a component has been destroyed. 

I originally was going to shirt circuit but that would have meant changing the return type as you wouldn’t have a cancelID any more.

scalvert:
Just circling back on this. 

> Because the point of these functions is to stop the developer having to check isDestroying and isDestroyed.

That's not really the primary point. The primary point is to auto-cancel any pending work that an object has scheduled _when_ it's destroyed. To address what you've described, the pattern we use is to assert. Unless there's a compelling reason to change this, I think this is the pattern we'll continue to use. If we want to change the behavior, we should change it for all functions, as they all behave the same (assert).

What do you think?

BryanCrotaz:
> auto-cancel any pending work that an object has scheduled _when_ it's destroyed

So we're saying that the developer doesn't need to worry that a task might execute after the component is destroyed for example. It is silently cancelled.

The same logic is true when setting up the task - it should be silently cancelled if the target is already destroyed.

This is all about making the app developer's life easy, and not having to check isDestroyed during the task code. I'm simply extending that to when scheduling it.

Otherwise I either have to write:
```
myDOMEvent() {
  if (!target.isDestroying && !target.isDestroyed) {
    scheduleTask(target, doTask)
  }
}

// or

myDOMEvent() {
  try {
    scheduleTask(target, doTask);
  } catch (e) {
    if (e is AssertionError) {
      // target has been destroyed, ignore error
    }
  }
}
```

In my suggestion the developer instead has to write:
```
myDOMEvent() {
  scheduleTask(target, doTask);
}
```

BryanCrotaz:
> we should change it for all functions, as they all behave the same (assert).

Yes, we should do this (but only for 'target is destroyed') - other assertions that you've passed valid data should stay. If they fail it's because of developer error, not runtime error.

scalvert:
Ya I understand the intent. Also, as an aside, I believe you only need to check `isDestroying` to determine if an object is destroyed, as it's a [bitwise flag that's set](https://github.com/emberjs/ember.js/blob/528434e88cb1b544ea4b965c81c6b2102466544f/packages/@ember/-internals/meta/lib/meta.ts#L185) and not unset.

We'd have to test for `destroyable.isDestroying` on each function call (*Task and DOM functions) on invocation. We shouldn't need to test for this on the actual scheduled function invocation, as those scheduled functions would have already been canceled when the object was destroyed. 

Is this what you're thinking? Also, I'd also not expect an empty function to be scheduled. Why wouldn't we not early return?

BryanCrotaz:
> We'd have to test for `destroyable.isDestroying` on each function call (*Task and DOM functions) on invocation

We're already doing this in an assert - I think the discussion is purely whether the target already being destroyed when the call is made is an error that should be thrown or silently ignored. My vote is for silently ignored.

> Also, I'd also not expect an empty function to be scheduled
I originally was going to short circuit but that would have meant changing the return type as you wouldn’t have a `cancelID` any more. This would mean the developer would have to check the return. yuk.

I suppose we could return -1 and check for this in `cancel`

scalvert:
Hey @BryanCrotaz, I'm going to close this in favor of #361. Thanks for submitting! 

I'd be interested in your opinion on that PR.

Issue #332, opened 2019-02-01
Typescript errors when running in host app


Gaurav0:
Upgraded my application to ember-cli-typescript@2.0.0-rc.2 from @1.5.0 and ember-cli-babel@7.4.1 from @6.18.0 and got the following errors:

```
ember s
node_modules/ember-lifeline/mixins/dom.d.ts:55:99 - error TS2304: Cannot find name 'RunMethod'.

55     addEventListener(this: MaybeIsComponent, target: TargetOrString, eventName: string, callback: RunMethod<MaybeIsComponent, any>, options?: Object | undefined): void;
                                                                                                     ~~~~~~~~~

node_modules/ember-lifeline/mixins/dom.d.ts:62:102 - error TS2304: Cannot find name 'RunMethod'.

62     removeEventListener(this: MaybeIsComponent, target: TargetOrString, eventName: string, callback: RunMethod<MaybeIsComponent, any>, options?: Object | undefined): void;
                                                                                                        ~~~~~~~~~
```

Gaurav0:
Minimal Reproduction: https://github.com/Gaurav0/repro-lifeline

buschtoens:
`RunMethod` is being referenced here:

https://github.com/ember-lifeline/ember-lifeline/blob/5ab67051e26f7f378a4cf3b698ce169743293bdc/addon/mixins/dom.ts#L68-L74

https://github.com/ember-lifeline/ember-lifeline/blob/5ab67051e26f7f378a4cf3b698ce169743293bdc/addon/mixins/dom.ts#L112-L118

It is defined here:

https://github.com/ember-lifeline/ember-lifeline/blob/5ab67051e26f7f378a4cf3b698ce169743293bdc/types/custom-ember.d.ts#L5-L7

Looking at this I wonder, if this ever worked in the first place? The `types` directory _is_ part of the npm tarball:

```bash
curl $(npm view ember-lifeline dist.tarball) | tar -ztf - 'package/types/*'
```

However, an addon's `types` directory is not included by the host app's `tsc`. You can "hotfix" this by adding `"node_modules/ember-lifeline/types/*"` to your `tsconfig.json` `"include"` array:

```bash
cat tsconfig.json \
  | jq '.include += ["node_modules/ember-lifeline/types/*"]' > tsconfig.json.fixed \
  && rm tsconfig.json \
  && mv tsconfig.json.fixed tsconfig.json
```

[...]

PR #336, opened 2019-02-06
fix typescript errors in host app

Gaurav0:
For #332


luketheobscure:
Just pointed my package.json at this commit, and all the errors went away. ❤️

rwjblue:
See https://github.com/ember-lifeline/ember-lifeline/issues/332#issuecomment-460018553

The correct fix is to move the custom types to be imported, instead of relying on the `types/**` folder "clobbering".

PR #345, opened 2019-02-18
Moves pollTaskFor out of poll-task into addon-test-support

scalvert:
@rwjblue as we've previously discussed, this moves `pollTaskFor` out of the `poll-task` module into `addon-test-support`. I was having some difficulty with re-exporting in `addon-test-support` for some reason, so right now we're importing directly from modules inside that directory. I'll work on figuring out what the issue is.


scalvert:
@rwjblue I think this is a pretty simple, straightforward change. I'm going to merge it in a few days if there's no specific feedback.

scalvert:
I agree. I'm going to pull a number of things together for a major bump.

PR #362, opened 2019-03-03
Updating project to latest addon blueprint

scalvert:


package.json
```diff
@@ -6,19 +6,25 @@
     "ember-addon"
   ],
   "license": "MIT",
-  "author": "Robert Jackson <me@rwjblue.com>",
+  "contributors": [
```
rwjblue: 👍

package.json
```diff
@@ -40,34 +47,35 @@
...
     "ember-cli": "~3.8.0",
-    "ember-cli-dependency-checker": "^3.0.0",
+    "ember-cli-dependency-checker": "^3.1.0",
     "ember-cli-htmlbars": "^3.0.0",
     "ember-cli-htmlbars-inline-precompile": "^2.0.0",
     "ember-cli-inject-live-reload": "^2.0.1",
-    "ember-cli-release": "^0.2.9",
-    "ember-cli-shims": "^1.1.0",
     "ember-cli-sri": "^2.1.0",
+    "ember-cli-template-lint": "^1.0.0-beta.1",
```
rwjblue: I'd _vaguely_ prefer to differ from the blueprint here, and just add `ember-template-lint` directly...

package.json
```diff
@@ -79,16 +87,16 @@
     "@types/jquery": "~3.3.22"
   },
   "engines": {
-    "node": "^4.5 || 6.* || >= 7.*"
+    "node": "6.* || 8.* || >= 10.*"
```
rwjblue: Just flagging this as a breaking change (again, totally fine with me).

Considering that we only have a month or two left in Node 6's support matrix, we should strongly consider making this:

```suggestion
    "node": "8.* || >= 10.*"
```

scalvert: I agree. I was planning on this being a major bump. I'd prefer to just jump up to 8+ and do a single major bump rather than two majors in quick succession. 

I'll go ahead with this unless there's any objections.

package.json
scalvert: Ya I'm fine with this.

scalvert:
Yes, I want to drop support for older versions of Ember. I'll update the README with both these suggestions.

package.json
rwjblue: No objections from me...

PR #374, opened 2019-03-07
Updates pollTaskFor to correctly await settled

### LINKED ISSUE #39: Maybe pollTaskFor should wait()?

### LINKED ISSUE #358: After Ember 3.2 upgrade, didRender() hook in a component with poll is not fired appropriately in qunit tests
I have been using ember lifeline's pollTask to animate a sliding progress bar component because it provides the easiest test interface for testing the different steps. The component has 5 segments that animate over different lengths of time by applying different CSS styles. In my tests, I test that after each `pollTaskFor` the appropriate CSS class has been applied.

When I upgrade my app from Ember 3.1 to Ember 3.2, this test starts to fail. It sees that the 1st and 3rd classes are applied, but not for 2nd, 4th, or 5th (I implemented a workaround to wait for settledness for the last one). After digging in a bit more, I see that the 2nd and 4th classes are NOT applied before the assert is run in the test, and it seems to be because the `didRender()` hook on the component is not called after these polls, but it is called after the 3rd. I don't understand why it is called sometimes and not other times, and why this would have changed when going from Ember 3.1 to Ember 3.2.

I pulled out the progress bar component into its own Ember app, and the test failure is still occurring in QUnit. I also tried upgrading the sample app in incremental minor version steps all the way up to 3.8, and the test failures never resolved. When I reverted back to Ember 3.1, the tests passed again.

You can see this app and download it yourself here: https://github.com/ajcolter/ember-progress-bar
I left a few branches open for the different versions of Ember. The master branch is on 3.2

scalvert:
Fixes #358 
Fixes #39


addon-test-support/poll-task-for.ts
```diff
@@ -1,12 +1,15 @@
 import { join } from '@ember/runloop';
 import { assert } from '@ember/debug';
-import { queuedPollTasks } from 'ember-lifeline';
+import { settled } from '@ember/test-helpers';
+import { queuedPollTasks, Token } from 'ember-lifeline';
 
-export default function pollTaskFor(token) {
+export default async function pollTaskFor(token: Token) {
```
rwjblue: 1. we don't use `await` so we dont' need `async` here
2. we can't rely on the host application to have regenerator runtime

```suggestion
export default function pollTaskFor(token: Token) {
```

tests/helpers/assert-async-throws.js
```diff
@@ -0,0 +1,17 @@
+import QUnit from 'qunit';
+
+QUnit.extend(QUnit.assert, {
+  asyncThrows: async function(func, regex, message = '') {
```
rwjblue: Umm, why?

Is there an issue with `assert.rejects`? 

https://api.qunitjs.com/assert/rejects

tests/unit/mixins/run-test.js
```diff
@@ -477,9 +475,13 @@ module('ember-lifeline/mixins/run', function(hooks) {
 
     assert.equal(calledTimes, 1, 'poll task argument was invoked initially');
 
-    assert.throws(function() {
-      pollTaskFor(token);
-    }, `You cannot advance pollTask '${token}' when \`next\` has not been called.`);
+    assert.asyncThrows(
```
rwjblue: should use `assert.rejects`

tests/unit/mixins/run-test.js
```diff
@@ -497,9 +499,13 @@ module('ember-lifeline/mixins/run', function(hooks) {
 
     run(component, 'destroy');
 
-    assert.throws(() => {
-      pollTaskFor(token);
-    }, `A pollTask with a token of ${token} was not found`);
+    assert.asyncThrows(
```
rwjblue: 🔪

tests/unit/mixins/run-test.js
```diff
@@ -524,9 +530,13 @@ module('ember-lifeline/mixins/run', function(hooks) {
 
     component.cancelPoll(token);
 
-    assert.throws(() => {
-      pollTaskFor(token);
-    }, new RegExp(`You cannot advance pollTask '${token}' when \`next\` has not been called.`));
+    assert.asyncThrows(
```
rwjblue: ✂️

tests/unit/poll-task-test.js
```diff
@@ -233,9 +233,13 @@ module('ember-lifeline/poll-task', function(hooks) {
 
     assert.equal(calledTimes, 1, 'poll task argument was invoked initially');
 
-    assert.throws(function() {
-      pollTaskFor(token);
-    }, `You cannot advance pollTask '${token}' when \`next\` has not been called.`);
+    assert.asyncThrows(
```
rwjblue: 💀

tests/unit/poll-task-test.js
```diff
@@ -254,9 +258,13 @@ module('ember-lifeline/poll-task', function(hooks) {
 
     cancelPoll(this.obj, token);
 
-    assert.throws(() => {
-      pollTaskFor(token);
-    }, new RegExp(`You cannot advance pollTask '${token}' when \`next\` has not been called.`));
+    assert.asyncThrows(
```
rwjblue: ☠️

PR #390, opened 2019-03-22
adding missing import

abeadam:
need to import runDisposables for the example to work right out of the box.


scalvert:
Thanks, @abeadam. Good catch.

PR #621, opened 2020-01-08
Dropping support for Node 8

scalvert:


rwjblue:
FWIW, we aren't deprecating we are dropping support.

scalvert:
Good clarification.

PR #852, opened 2020-07-20
Convert internal destroyable implementation to use @ember/destroyable

scalvert:
Implements https://github.com/ember-polyfills/ember-destroyable-polyfill/issues/8

- Converts lifeline's internal disposables implementation to use `ember-destroyable-polyfill`. 
- Removed the ability to `runDisposables` multiple times without destroying the underlying object
- Added deprecations to `registerDisposable` and `runDisposables`
- Added deprecations to all mixin implementations
- Changed `setupLifelineValidation` now encompasses all types of underlying object (including framework classes that are not using lifeline)

Follow up PR will...
1. Convert types to align better with built-ins (`Map` -> `Record`, `IDestroyable` -> `Destroyable`)


scalvert:
cc/ @buschtoens

addon/debounce-task.ts
```diff
@@ -1,7 +1,8 @@
 import { assert } from '@ember/debug';
 import { cancel, debounce } from '@ember/runloop';
 import { IDestroyable, IMap, EmberRunTimer } from './types';
-import { registerDisposable } from './utils/disposable';
+import { registerDestructor } from 'ember-destroyable-polyfill';
```
rwjblue, line 4: Should be imported from the official location (not the polyfill module):

```suggestion
import { registerDestructor } from '@ember/destroyable';
```

addon/debounce-task.ts
```diff
@@ -1,7 +1,8 @@
...
+import { Destructor } from 'ember-destroyable-polyfill/-internal/destructors';
```
rwjblue, line 5: O_o

scalvert:
cc/ @pzuraq

addon/dom-event-listeners.ts
```diff
@@ -1,6 +1,7 @@
 import { assert } from '@ember/debug';
 import { bind } from '@ember/runloop';
-import { registerDisposable } from './utils/disposable';
+import { registerDestructor } from 'ember-destroyable-polyfill';
```
rwjblue, line 3:
```suggestion
import { registerDestructor } from '@ember/destroyable';
```

addon/dom-event-listeners.ts
```diff
@@ -1,6 +1,7 @@
...
+import { Destructor } from 'ember-destroyable-polyfill/-internal/destructors';
```
rwjblue, line 4: Don't think we should need (or use) this private internal interface (it is not in the RFC)...

scalvert:
I'm not 100%, but I _think_ we may be able to remove the need, internally, for the `WeakMap`s to track the metadata used for cancelation. I think we can simply register a destructor that closes over those values. 

@pzuraq, @rwjblue do you know?

addon/poll-task.ts
```diff
@@ -2,8 +2,9 @@ import Ember from 'ember';
 
 import { deprecate } from '@ember/application/deprecations';
 import getTask from './utils/get-task';
-import { registerDisposable } from './utils/disposable';
 import { IMap, TaskOrName, IDestroyable } from './types';
+import { registerDestructor } from 'ember-destroyable-polyfill';
```
rwjblue, line 6:
```suggestion
import { registerDestructor } from '@ember/destroyable';
```

addon/poll-task.ts
```diff
@@ -2,8 +2,9 @@ import Ember from 'ember';
...
+import { Destructor } from 'ember-destroyable-polyfill/-internal/destructors';
```
rwjblue, line 7: 🔪

addon/run-task.ts
```diff
@@ -9,7 +9,8 @@ import {
   EmberRunTimer,
   EmberRunQueues,
 } from './types';
-import { registerDisposable } from './utils/disposable';
+import { registerDestructor } from '@ember/destroyable';
+import { Destructor } from 'ember-destroyable-polyfill/-internal/destructors';
```
rwjblue, line 13: ✂️

tests/unit/debounce-task-test.js
```diff
@@ -15,10 +15,6 @@ module('ember-lifeline/debounce-task', function (hooks) {
     };
   });
 
-  hooks.afterEach(function () {
-    runDisposables(this.obj);
```
rwjblue, line 19: We still need to do this conceptually, but its probably `import { destroy } from '@ember/destroyable'; destroy(this.obj);`

tests/unit/dom-event-listeners-test.js
```diff
@@ -26,12 +26,6 @@ module('ember-lifeline/dom-event-listeners', function (hooks) {
           this._super(...arguments);
           testContext.componentInstance = this;
         },
-
-        destroy() {
-          this._super(...arguments);
-
-          runDisposables(this);
```
rwjblue, line 33: I think this is good to remove, we should no longer _need_ to call `runDisposables` manually (since Ember / Glimmer do this for us).

We should add some tests that confirm things work fine when you **do** still call it manually though.

scalvert:
It appears we need to determine a path forward WRT the latest version we want to support here. 3.4 is out of LTS, so we should consider removing.

rwjblue:
> I think we can simply register a destructor that closes over those values.

Ya, we can, but then we'd either have to always schedule a new destructor for each call to `runTask` and I don't know that that makes sense in context.

addon/run-task.ts
scalvert: Err. Cut which? We need the destroyable import.

tests/unit/debounce-task-test.js
scalvert: Hmmm. OK. I can update them.

tests/unit/dom-event-listeners-test.js
scalvert: Agree.

addon/debounce-task.ts
scalvert: Oh alskdjflaskdf. I think that was vscode doing that shenanigans.

addon/debounce-task.ts
scalvert: DON'T LOOK!

addon/dom-event-listeners.ts
scalvert: Hmmm ok. I _do_ need to type the second param to `registerDestructor`. Let me see how I can remove it.

tests/unit/dom-event-listeners-test.js
scalvert: To be clear, I can add separate tests that test the disposables APIs themselves.

addon/dom-event-listeners.ts
scalvert: I removed it.

tests/unit/dom-event-listeners-test.js
rwjblue: > I can add separate tests that test the disposables APIs themselves.

Yes, exactly. In that set of tests we **must** ensure that calling `runDestroyables` _from `willDestroy`_ does not break the system, it is a bit nuanced but IMHO quite important.

.github/workflows/ci-build.yml
```diff
@@ -48,7 +48,6 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          - ember-lts-3.4
```
rwjblue, line 51: This seems unrelated, does it cause failures with this API? Ideally we do not need to change our support constraints in order to migrate to the new public API...

tests/unit/dom-event-listeners-test.js
```diff
@@ -557,37 +531,4 @@ module('ember-lifeline/dom-event-listeners', function (hooks) {
     await triggerEvent(window, 'click');
     assert.equal(calls, 1, 'callback was called');
   });
-
-  test('runDisposables more than once', async function (assert) {
```
rwjblue, line 561: I think we need to bring this test back and add another test for calling runDestroyables _while_ the native / platform destroyables are running.

scalvert:
Seems fair. At first glance the 3.4 issue seems to be related to changed internal paths that the polyfill relies on. We can for sure update the polyfill to be more resilient to these changes. I'll dig in and, if needed, open a PR.

rwjblue:
> We can for sure update the polyfill to be more resilient to these changes. I'll dig in and, if needed, open a PR.

Ya, even just an issue over there (or a PR adding the ember-try scenario). It is possible there is a concrete reason it doesn't work on 3.4 (some feature that was added that fundamentally unlocks it) but I can't think of anything...

tests/unit/dom-event-listeners-test.js
scalvert: Agree.

scalvert: We discussed offline and agreed to not add this, since effectively this is _not_ expected to work when this is invoked multiple times with the new APIs.

scalvert: The `runDisposables` function should effectively be a noop, as the system is now responsible for running the destructors.

.github/workflows/ci-build.yml
scalvert: Removed support for 3.4 in this PR https://github.com/ember-lifeline/ember-lifeline/pull/855

tests/unit/dom-event-listeners-test.js
scalvert: As per my previous comment. We're removing this feature (running `runDestroyables` multiple times).

rwjblue:
FYI - I landed fixes for Ember 3.4 and Ember 3.20 over in `ember-destroyables-polyfill`, but still needs a release (need to get added to owners on npm still).

rwjblue:
OK, https://github.com/ember-polyfills/ember-destroyable-polyfill/releases/tag/v1.0.0 is out. It fixes both 3.20 and 3.4/3.8.

@scalvert mind updating?

scalvert:
No problem. Will do that today. Thanks for the fixes!

tests/unit/dom-event-listeners-test.js
scalvert: Looking at this a bit closer - can you explain your rationale for wanting these tests? The reason I ask is that, without any disposables being registered for a given object, the `runDisposables` call would early return. What are we trying to test here?

I still believe that adding this specific test - running disposables more than once - should be added to the core of the disposables utility tests.

.github/workflows/ci-build.yml
rwjblue: Will need to be rebased, since this already landed

tests/unit/dom-event-listeners-test.js
scalvert: Synced offline. We determined the correct number of tests we need.

addon/mixins/disposable.ts
```diff
@@ -1,5 +1,5 @@
 import Mixin from '@ember/object/mixin';
```
rwjblue, line 1: Can you add a deprecation to this mixin (likely by adding an `init` and issuing the deprecation)?

scalvert: Yes. I was thinking we should deprecate them along with this too.

rwjblue: Want to do that here, or in a separate PR?

scalvert: Done

scalvert:
Updated the description to contain the list of changes. Thanks for the back-and-forth, @rwjblue!

Issue #876, opened 2020-08-05
Parameter type error


nightire:
https://github.com/ember-lifeline/ember-lifeline/blob/8ade5a78bcb3cd6a6849277e18ebcf5bfae2f057/addon/run-task.ts#L276

The `cancelId` should be typed as `Timer` because it can be a primitive number as well.

nlfurniss:
It's can't because `Ember.runloop.cancel` wants an `EmberRunTimer`, so it can't be a number

PR #877, opened 2020-08-05
fix(#876): parameter type error

nightire:


addon/run-task.ts
```diff
@@ -273,7 +273,7 @@ export function throttleTask(
  */
 export function cancelTask(
   destroyable: IDestroyable,
-  cancelId: EmberRunTimer
+  cancelId: Timer
```
rwjblue, line 276: Ultimately this should be the (somewhat opaque) result of calling `runTask`. As long as our internal types agree, I'm not sure what changing this helps. Can you help me understand under which circumstances you are getting something typed as `number` that you need to pass in here?

scalvert: Support for both `EmberRunTimer` _and_ `number` (what this `Timer` type alias does) was explicitly removed as part of the deprecation that was targeted for v4.0.0 (https://github.com/ember-lifeline/ember-lifeline/pull/859/files#diff-83f8a6c941a5fd15457b4d5ce99e52f2R276). 

I don't believe there's a valid use case to restore this, but if there is please let us know.

nightire: My use case is `runTask` returns a `Timer`, not `EmberRunTimer`, maybe we should fix the return type of `runTask` instead?

nightire: > Ultimately this should be the (somewhat opaque) result of calling `runTask`. As long as our internal types agree, I'm not sure what changing this helps. Can you help me understand under which circumstances you are getting something typed as `number` that you need to pass in here?

@rwjblue @scalvert Here returns a number:

https://github.com/ember-lifeline/ember-lifeline/blob/472de2c188e0879cbe2f6abc9450b6310ae46bbc/addon/run-task.ts#L80-L82

Thus, `runTask` returns a `cancelId` that possible to be a number. Currently I can't pass this `cancelId` to the `calcelTask` function because it will cause a type error.

@rwjblue @scalvert

PR #931, opened 2020-09-17
Upgrades ember-destroyable-polyfill to fix issue with double _super invocation

scalvert:
Pulls in fix for https://github.com/ember-polyfills/ember-destroyable-polyfill/pull/112


rwjblue (approved):
Seems good to bump and release, just to make sure folks don't get trolled by the issue that 2.0.2 was working around.

PR #980, opened 2021-04-29
Upgrade caniuse-db and dedupe yarn.lock

nlfurniss:
Prompt said it's out-of-date so now it's not!


scalvert:
Thank you kind sir!

PR #988, opened 2021-05-03
Adds deprecation warning for DOM APIs

scalvert:
Targets removal for 7.0.0.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -39,7 +39,6 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          - ember-lts-3.8
           - ember-lts-3.12
           - ember-lts-3.16
           - ember-lts-3.20
--- a/addon/dom-event-listeners.ts
+++ b/addon/dom-event-listeners.ts
@@ -104,6 +104,11 @@ export function addEventListener<T extends IDestroyable>(
     {
       id: 'ember-lifeline-deprecated-addeventlistener',
       until: '7.0.0',
+      // @ts-ignore
+      for: 'ember-lifeline',
+      since: {
+        enabled: '6.0.0',
+      },
     }
   );
 
@@ -161,6 +166,11 @@ export function removeEventListener<T extends IDestroyable>(
     {
       id: 'ember-lifeline-deprecated-removeeventlistener',
       until: '7.0.0',
+      // @ts-ignore
+      for: 'ember-lifeline',
+      since: {
+        enabled: '6.0.0',
+      },
     }
   );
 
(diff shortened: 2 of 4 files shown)


addon/dom-event-listeners.ts
```diff
@@ -98,6 +98,15 @@ export function addEventListener<T extends IDestroyable>(
 ): void {
   assertArguments(target, eventName, callback);
 
+  deprecate(
+    'ember-lifeline addEventListener is deprecated. Please use modifiers instead.',
+    false,
+    {
+      id: 'ember-lifeline-deprecated-addeventlistener',
+      until: '7.0.0',
+    }
```
rwjblue, line 107: We also need to add `since` and `for` properties (if they aren't present, ember-source@3.27+ will issue a deprecation).

See example commit https://github.com/emberjs/ember-test-helpers/commit/a7d7e3e664572a10d7e44f6ac4974a63df9070cf

addon/dom-event-listeners.ts
```diff
@@ -146,6 +155,15 @@ export function removeEventListener<T extends IDestroyable>(
 ): void {
   assertArguments(target, eventName, callback);
 
+  deprecate(
+    'ember-lifeline removeEventListener is deprecated. Please use modifiers instead.',
+    false,
+    {
+      id: 'ember-lifeline-deprecated-removeeventlistener',
+      until: '7.0.0',
+    }
```
rwjblue, line 164: Same thing here, `since`/`for`

.github/workflows/ci-build.yml
```diff
@@ -39,7 +39,6 @@ jobs:
       fail-fast: true
       matrix:
         ember-try-scenario:
-          - ember-lts-3.8
```
rwjblue, line 42: Why do we need to drop this?

rwjblue (changes requested):

.github/workflows/ci-build.yml
scalvert: Again, as I mentioned to you via chat, this scenario was hanging, and rather than my worrying about fixing it (since we're removing support for this in another PR), I'd rather just remove it.

addon/dom-event-listeners.ts
scalvert: Added.

addon/dom-event-listeners.ts
scalvert: Added.

.github/workflows/ci-build.yml
rwjblue: Gotcha. Yeah, its fairly bad but since this PR itself is definitely not breaking anything I think that is fine.

rwjblue (approved):

PR #990, opened 2021-05-05
Cleanup, deprecation removal, and blueprint updating.

scalvert:
Updates blueprint to `3.27.0-beta.1.`

- Updates dummy app to use Octane syntax
- Updates tests to remove EmberObject usage
- Removes deprecated APIs slated for removal in 7.0.0.

🚨 breaking changes

- Drops support for Node 10
- Removes disposable APIs (targeted for removal in 7.0.0)
- Removes DOM APIs (targeted for removal in 7.0.0)


package.json
```diff
@@ -44,12 +49,16 @@
     ]
   },
   "dependencies": {
-    "ember-cli-babel": "^7.23.0",
+    "ember-cli-babel": "^7.26.5",
+    "ember-cli-htmlbars": "^5.7.1",
```
rwjblue, line 53: We don't have any `addon/**/*.hbs` files, mind moving this back down into devDeps?

rwjblue (commented):
Looks good! I left one inline comment RE: ember-cli-htmlbars, but I also think we should update the README a tad bit:

* Call out minimum Ember / ember-cli support of 3.16
* Call out that we are following the browser support matrix defined by [emberjs/rfcs#685](https://github.com/emberjs/rfcs/blob/master/text/0685-new-browser-support-policy.md) (and explicitly dropping IE11 support)

Additionally, I'm wondering if we should also leverage the V2 addon shim (since this will be a major bump anyways). What do you think?

rwjblue:
https://www.npmjs.com/package/@embroider/addon-shim is the addon shim, the changes are pretty small (to our `index.js` at least)...

rwjblue:
Pushed the changes that I mentioned, this should be good to land once CI is happy.

PR #1096, opened 2022-02-05
Update docs to use Octane throughout

chriskrycho:
As a bonus: get rid of the now-long-defunct mixin docs! :tada:

### DIFF
--- a/packages/test-app/app/templates/docs/poll-task.md
+++ b/packages/test-app/app/templates/docs/poll-task.md
@@ -17,22 +17,23 @@ and performance problems. Instead, you should be scheduling new work
 _after_ the previous work was done. For example:
 
 ```js
-import Component from '@ember/component';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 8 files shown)


packages/test-app/app/templates/docs/poll-task.md
```diff
@@ -75,29 +77,30 @@ This is where `pollTask` really shines. You could rewrite the above example to u
 like this:
 
 ```js
-import Component from '@ember/component';
-import { inject } from '@ember/service';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { service } from '@ember/service';
```
chriskrycho, line 82: Note: this assumes Ember v4. Seems fine?

scalvert: I think so. Doesn't mean folks can't use it on older versions.

scalvert:
Thanks for doing this, @chriskrycho!

scalvert (approved):

PR #1107, opened 2022-04-11
Extracting utils from run-task

scalvert:
For some reason I can't remember, the `runTask`, `scheduleTask` and `throttleTask` utils were all contained in the same file. The `debounceTask` util was already split out into its own file (cuz it's special, apparently). 

This change splits the other utils to their own separate files, making it easier to grok where everything lives.


scalvert:
Going to merge this through as it's largely a structural change, not functional.

PR #1117, opened 2022-04-25
Renaming types and defaulting to Ember types

scalvert:
This PR attempts to clean up our types a bit, defaulting to Ember types over internally created types. It also renames interface names to more conventional names (without the C#-inspired "I" prefix).


scalvert:
Just going to merge this in as it's largely renames, albeit will be breaking for TS consumers if they use our types.

PR #1137, opened 2022-09-16
Created timer util that returns a scaled-down timer value in test environments.

tgvrssanthosh:
## Summary

## Details

### DIFF
--- a/addon/utils/timer.ts
+++ b/addon/utils/timer.ts
@@ -1,0 +1,18 @@
+import Ember from 'ember';
+
+/**
+ * @function getTestAwareTimerVal
+ * @param {number} timerVal The timer value that gets scaled down in case of the test environment.
+ * @param {number} [customTestTimerVal] The timer value that needs to be set in the test environment if passed
+ */
+export function getTestAwareTimerVal(
+  timerVal: number,
+  customTestTimerVal?: number
+): number {
+  if (Ember.testing) {
+    return customTestTimerVal !== undefined
+      ? customTestTimerVal
+      : timerVal / 100; // Reducing timer value by 100x in test environment
+  }
+  return timerVal;
+}
--- a/addon/utils/timer.ts
+++ b/addon/utils/timer.ts
@@ -1,18 +1,22 @@
 import Ember from 'ember';
+import { TestTimeoutOptions } from '../types';
 
 /**
- * @function getTestAwareTimerVal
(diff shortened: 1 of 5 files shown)


addon/utils/timer.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param {number} [customTestTimerVal] The timer value that needs to be set in the test environment if passed
+ */
+export function getTestAwareTimerVal(
+  timerVal: number,
+  customTestTimerVal?: number
+): number {
+  if (Ember.testing) {
+    return customTestTimerVal !== undefined
+      ? customTestTimerVal
+      : timerVal / 100; // Reducing timer value by 100x in test environment
```
scalvert, line 15: I wonder if we should be passing this value too? I can imagine consumers wanting to specify the scaling value themselves.

Perhaps we should provide the second argument as an options object, where we could have either an explicit timer value, or a scaling factor.

e.g.

```ts
interface TestTimeoutWithTimeout {
  timeout: number;
  scaling?: never;  
}
interface TestTimeoutWithScaling {
  timeout?: never;
  scaling: number;
}
type TestTimeoutOptions = TestTimeoutWithTimeout | TestTimeoutWithScaling;
```

I also think we should mimic the naming from the actual utils themselves. Specifically, we call that [argument `timeout` in `runTask`](https://github.com/ember-lifeline/ember-lifeline/blob/master/packages/ember-lifeline/src/run-task.ts#L39). I think we should call it the same, since that's the value we expect it to represent.

addon/utils/timer.ts
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+/**
+ * @function getTestAwareTimerVal
+ * @param {number} timerVal The timer value that gets scaled down in case of the test environment.
+ * @param {number} [customTestTimerVal] The timer value that needs to be set in the test environment if passed
+ */
+export function getTestAwareTimerVal(
```
scalvert, line 8: Can we rename this?

```suggestion
export function getTimeoutForTesting(
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-lifeline",
-  "version": "6.0.2",
```
scalvert, line 3: Can we revert this? We use `release-it` to bump versions, and that process will auto-bump the version number.

scalvert (changes requested):

PR #1140, opened 2022-09-21
Created timer util that returns a scaled-down timer value in test environments.

tgvrssanthosh:
## Summary

## Details


packages/ember-lifeline/src/utils/timer.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param {number} timeoutVal The timeout value that gets scaled down in case of the test environment.
+ * @param {number} [customTestTimeout] The timeout value that needs to be set in the test environment if passed
+ */
+export function getTimeoutOrTestFallback(
+  timeoutVal: number,
+  options?: TestTimeoutOptions
+): number {
+  if (Ember.testing) {
+    const customTestTimeout = options?.timeout;
+    const scaleDownFactor = options?.scaling ?? 100; // Reducing timer value by 100x in test environment by default
```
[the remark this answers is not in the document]
scalvert: Also, can we just call it `scaling`?

```suggestion
    const scaling = options?.scaling ?? 100; // Reducing timer value by 100x in test environment by default
```

packages/ember-lifeline/src/utils/timer.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @function getTimeoutOrTestFallback
+ * @param {number} timeoutVal The timeout value that gets scaled down in case of the test environment.
+ * @param {number} [customTestTimeout] The timeout value that needs to be set in the test environment if passed
+ */
+export function getTimeoutOrTestFallback(
+  timeoutVal: number,
+  options?: TestTimeoutOptions
+): number {
+  if (Ember.testing) {
+    const customTestTimeout = options?.timeout;
```
scalvert, line 14: Can we just call this `timeout`?

```suggestion
    const timeout = options?.timeout;
```

scalvert (changes requested):

packages/ember-lifeline/src/utils/timer.ts
tgvrssanthosh: Since there is another param named `timeout`, I couldn't rename it to timeout. Instead, i have renamed it to `testTimeout`.

scalvert (approved):

PR #1145, opened 2022-09-23
Back ported the timer util that returns a scaled down timer value in test environments

tgvrssanthosh:
## Summary

## Details


scalvert (approved):

package.json
```diff
@@ -97,13 +99,14 @@
...
-    "typescript": "~4.0.3"
+    "typescript": "~4.0.3",
+    "webpack": "^5.68.0"
   },
   "resolutions": {
     "favicons": "5.3.0"
   },
   "engines": {
-    "node": ">= 10.*"
+    "node": "12.* || 14.* || >= 16"
```
scalvert, line 109: We can't make this change in the v6 branch. Changing node engines is a breaking change, and as such it's not possible to do.

package.json
```diff
@@ -61,8 +62,9 @@
     "@typescript-eslint/parser": "^3.10.1",
     "broccoli-asset-rev": "^3.0.0",
     "documentation": "^13.0.2",
+    "ember-auto-import": "^2.4.2",
     "ember-cli": "~3.21.2",
-    "ember-cli-addon-docs": "^0.8.0",
+    "ember-cli-addon-docs": "^4.2.2",
```
scalvert, line 67: I'm not sure we should be bumping addon-docs all the way up to this version. I'd prefer we do minimal changes in this release branch, and defer changes to the master branch.

scalvert (changes requested):

elwayman02:
@tgvrssanthosh can you update this PR according to the review feedback, to avoid introducing breaking changes?

elwayman02:
Please rebase now that #1149 is merged.

PR #1149, opened 2022-10-13
[backport] add missing fields (for, since) to deprecate method for ember 4 compat

hmajoros:
Fixes the following deprecations to unblock folks on v6 of the library trying to use Ember 4.x

https://deprecations.emberjs.com/v3.x/#toc_ember-source-deprecation-without-for
https://deprecations.emberjs.com/v3.x/#toc_ember-source-deprecation-without-since

### DIFF
--- a/addon/mixins/disposable.ts
+++ b/addon/mixins/disposable.ts
@@ -20,6 +20,10 @@ export default Mixin.create({
       {
         id: 'ember-lifeline-deprecated-disposable-mixin',
         until: '7.0.0',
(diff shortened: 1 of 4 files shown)


rwjblue (approved):

scalvert:
CI has the same issue as in #1145.

scalvert:
In fact, these changes are already present in #1145, so I'm not sure we should duplicate. We should make these changes in either this PR or that, but not both.

PR #1174, opened 2023-05-04
Ember 4 Upgrade

elwayman02:
## Summary

Updates for v7 to support Ember@4

Internal:
* eslint 8
* Swap node lint plugin for replacement
* Fix assert.equal lint failures
* Run prettier
* Remove destroyable polyfill (inert after 3.22)
* Drop Ember 3.24 support
* Add Ember 4.x test scenarios
* Add missing routes for new docs (added in previous PRs)
* Drop yuidocs plugin (unused)
* Bump various ember-related deps to latest
* Remove 2 archived/unneeded ember deps
* Add required @ember/string dep
* Drop support for Node 10/12/14
* Bump release-it deps to latest


elwayman02:
Embroider test failures are due to needing to add ember-data as a devDep of the test app because ember-cli-addon-docs requires it as a peerDep, and recent changes cause the demo app to break without it:

```
Uncaught Error: While generating link to route "docs": Assertion Failed: Attempting to inject an unknown injection: 'service:store'
```

It's highly likely that code in the addon package is still embroider-compatible, but because the demo app and test app are tightly coupled, we cannot have a passing test run.

We should be ok to merge this for now, and the long-term fix would be to split the entire demo app into its own package within the workspace, such that the test app no longer needs dependencies related to ember-cli-addon-docs.

elwayman02:
To account for this, I've marked the embroider builds as "experimental" in the GHA config for now.

scalvert (approved):

Issue #1177, opened 2023-09-14
Type imports only available in dist/ and src/


SanderKnauff:
After updating to 7.0.0. I had to update some imports to include either `dist` or `src` in the import path:

```ts
- import { TaskOrName } from 'ember-lifeline/types';
+ import { TaskOrName } from 'ember-lifeline/dist/types';
```

While this does work, it feels a bit dirty. Was this change done intentionally?

Issue #1176, opened 2023-05-15
test-support types not found after upgrade to 7.0.0


dagroe:
After upgrading to `7.0.0` I get a TypeScript error in tests when importing from `ember-lifeline/test-support`.

E.g. in `tests/integration/components/component-1-test.ts` I get error 

`Cannot find module 'ember-lifeline/test-support' or its corresponding type declarations.`

for import

```typescript
import { pollTaskFor } from 'ember-lifeline/test-support';
```

Do I need to change those imports?

PR #1181, opened 2024-11-25
Export types from ./types

dagroe:
It seems this export is missing so that one can also import types like `Destoryable` from `ember-lifeline`, e.g. `
import { type Destroyable } from 'ember-lifeline';`

This likely solves issues #1176 and #1177


SanderKnauff:
@rwjblue @scalvert apologies for the ping, but I noticed that you two were owners of the org.

I ran into this problem as well a few weeks ago. Is this something that could be merged?

Issue #1182, opened 2026-04-05
Breaks with ember-source 7 beta


TSenter:
I have tests that run almost weekly against ember-source alpha and beta. Since `ember-source@7.0.0-alpha.1` and `ember-source@7.0.0-beta.1` have now been published as the latest alpha and beta versions, the usage of the Ember barrel file no longer works.

PR #1183, opened 2026-04-13
refactor!: Use Embroider macro in favor of Ember barrel file

TSenter:
The latest beta release of Ember is 7.0.0, which removes the barrel file used here. We can switch to the `@embroider/macros` library to check for testing, although it should probably be considered a breaking change.


johanrd:
@TSenter thanks! this looks good. currently a blocker for upgrade to the newly released ember-source `v7.0.0`.

TSenter:
Resolves #1182

NullVoxPopuli (approved):

